PR #324, opened 2023-11-09
fix for failling tests when OS uses non-US locale

xxBedy:


novoj:
Thank you very much @xxBedy !

PR #438, opened 2024-01-24
Metrics

Khertys:
This PR contains the current state of observability API. Tomorrow, there will be hopefully the last refactor before merging into dev.

Any ideas with improvements would be great.

@lukashornych Valid handling of `requestId` and `clientId` in GraphQL handlers is one thing that is not yet done. At the moment I'm just passing the `traceId` value parsed from the OpenTelemetry header into the `requestId`, while ignoring the passed and still sent value in the `request_id` header. Shall we keep this header or do we remove it?


novoj:
I did some minor changes and here are my remarks:

- [ ] Can you please document:
   - [ ]  `io.evitadb.externalApi.trace.OpenTelemetryTracerSetup`
   - [ ] `io.evitadb.externalApi.utils.JsonApiClientContext`
   - [ ] `io.evitadb.externalApi.utils.ProtobufApiClientContext`
   - [ ] `io.evitadb.externalApi.observability.ObservabilityManager` (class level)
- [ ] Do we need two scopes here? `io.evitadb.externalApi.utils.ExternalApiClientContext#executeWithClientAndRequestId(java.net.SocketAddress, java.lang.String, java.lang.String, java.lang.Runnable)` ... shouldn't the clientId be a span attribute?
- [ ] could be `traceparent` wrapped in some kind of constraint? `io/evitadb/externalApi/graphql/GraphQLManager.java:254`
- [ ] shouldn't we throw exceptions here `io.evitadb.externalApi.observability.ObservabilityManager#stop` and translate them to HTTP STATUS ERROR CODES to provide proper REST API?! I can't imagine how would client parse text error messages ...

lukashornych:
Most of it is just stylistic stuff or things we've already talked about. Overall it looks very good.

- [ ] there are missing `@Nonnull` and `@Nullable` annotations in some places (e.g. `MetricHandler`)
- [ ] i would move the `UsedMetric` class into evita_api, it would be used by potentially all modules right?
- [ ] why there is `requires jdk.httpserver;` in `evita_engine/module-info.java`? Seems odd
- [ ] shouldn't the `OpenTelemetryTracerSetup` by in the observability module as we talked about?
  - [ ] also, i feel like the `getGrpcTelemetry` method should be separate in the grpc server module that would return the interceptor in one go
  - [ ] similarly, maybe move the `extractContextFromHeaders` to the core into the `JsonApiClientContext`?
- [ ] i would remove the `requestId` from `ExternalApiClientContext` as it is passed from external context already as `traceId`
- [ ] i would maybe move the `ProtobufApiClientContext` to the grpc server module as it is not something common for multiple modules
  - [ ] maybe it could be merged into the `GrpcClientContext`
- [ ] `evita_external_api_code/module-info.java` contains OTEL requires (just so it is logged that we wanted to do this)
  - [ ] i think the `opentelemetry-bom` is similar, could it be in the observality module only?
- [ ] in `GraphQLHandler`/`RestEndpointHandler` i would maybe pass the entire`serverExchange` into the clientContext instead of separate properties
- [ ] the `traceparent` header could be added to the existing `AdditionalHeaders` class in core
- [ ] old class docs in `ObservabilityConfig`
- [ ] the `LoggingEndpointHandler` should ideally use observability-specific exceptions, not the base external api ones
- [ ] seems odd that the `StartLoggingHandler` and `StopLoggingHandler` are returning 404 if there is an error, you shouldn't need try..catch for this, that's what the exception handler is for, the only time you should need the try..catch is in the manager (start/stop) to wrap non-evitaDB exception to evitaDB custom exception (probably your own)
  - [ ] e.g. the `stop` method should throw appropriate custom exceptions with the error messages, not returned direct messages, and should wrap the IOException to a custom exception as well, the returned string should contain only filename (this would also result in inconsistent HTTP statuses)
- [ ] `RegisteredCustomEventProvider` could be package-private
- [ ] to be honest i don't understand the idea behind pre-defined `eventPackageNames` in the `CustomEventProvider`, shouldn't this be the place where the packages are resolved automatically from class list?
- [ ] `MetricHandler#116` the condition seems odd, you use `isCustomEventClassName` here same as on `#134` but here is it results in JVM metrics, while on the `#134`  it results in custom evita events?
- [ ] `MetricHandler#138` seems odd, shouldn't there be `else`? 
- [ ] in `MetricHandler` should you use cached `ReflectionLookup` ?
- [ ] the annotation parsing should be done once and result in some `descriptor` classes or something instead of doing it over and over in `MetricHandler`
- [ ] in `MetricHandler` why is the `eventMetricClasses` a Map if you only need its values? It could be all simpler if you just create single Set of registered classes, right?
- [ ] the `ObservabilityManager` should use custom internal errors as well

+ the @novoj notes

novoj:
Regarding:

> in MetricHandler should you use cached ReflectionLookup

This was discussed and because it's "one-time" registration there is no benefit of using cached access - on the contrary we would consume memory, that will not bring us any benefit in the future (we will not ask again for the data).

PR #644, opened 2024-08-12
fix: lab gui handler windows fix - path standardized inside of jar to forward slashes on all systems si

Khertys:
lab gui handler windows fix - path standardized inside of jar to forward slashes on all systems si


novoj:
@lukashornych is if completely fixed? Or should I track this problem further?

lukashornych:
@Khertys told me that it doesn't have a better solution, so I take it as solved.

PR #769, opened 2025-01-07
fix: facetXXXGroupsXXXjunction constraint target reference entity type instead of reference group type

### LINKED ISSUE #757: Invalid GraphQL schema for facetGroup conjunction/negation/disjunction
The schema offers attributes of the `tag` entity instead of `tagCategory` entity:

image

These constraints focus directly on reference schema group and need to offer properties of this schema instead.

lukashornych:
Resolves #757


novoj:
I've retested the problematic scenario and it seems to be ok, thanks! 😍

Issue #1257, opened 2026-06-26
Sort by multi-valued referenceProperty with a soft-deleted reference attribute returns duplicate PKs (and crashes on reference fetch)


novoj:
## Summary

Queries that sort by `referenceProperty` on a multi-valued reference and fetch reference content can crash with:

```
java.lang.IllegalStateException: Duplicate key <PK>
  at java.util.stream.Collectors.duplicateKeyException(Collectors.java:135)
  at io.evitadb.core.query.fetch.EntityIndexSupplier.get(EntityIndexSupplier.java:59)
  at io.evitadb.core.query.fetch.ReferencedEntityFetcher.lambda$initReferenceIndex$46(ReferencedEntityFetcher.java:1923)
  ...
```

The crash is the proximate symptom. The underlying defect is **duplicate primary keys in the sorter output**: even without `entityFetch(referenceContent(...))`, the same query returns a page whose `int[]` of primary keys contains the same entity twice. The duplicate-key `Collectors.toMap` in `EntityIndexSupplier` is the policy-correct fail-loud that exposes it.

## Root cause

Two attribute-access paths in the sort code use raw `Attributes.getAttribute(...)`, which returns the value of a soft-deleted attribute (`AttributeValue.dropped() == true`) as if it were live:

- `evita_engine/src/main/java/io/evitadb/core/query/sort/attribute/comparator/AbstractReferenceAttributeComparator.java:131`
  — `final Serializable storedAttributeValue = it.getAttribute(attributeName);`
- `evita_engine/src/main/java/io/evitadb/core/query/sort/attribute/comparator/AbstractReferenceCompoundAttributeComparator.java:125-127`
  — `this.attributeValueFetcher = locale == null ? AttributesContract::getAttribute : (rc, an) -> rc.getAttribute(an, locale);`

Every other surface honours `dropped()`: `ReferenceDecorator` filters via its `attributePredicate`, the REST/GraphQL fetch correctly omits dropped attributes, and the `SortIndex` rebuild excludes them — so a soft-deleted reference-attribute is **absent** from `SortedRecordsProvider.indexOf(pk)`. The comparator's raw fetcher disagrees: it sees the historical Predecessor value and treats the entity as sortable.

That disagreement drives the failure in `TraverseReferencePredecessorAttributeComparator.compare` (`evita_engine/.../comparator/TraverseReferencePredecessorAttributeComparator.java:172-272`). For an entity with a dropped sortable attribute:

[...]

PR #1260, opened 2026-06-26
chore: untrack agent reproduction artifacts (follow-up to #1258)

novoj:
Follow-up to #1258 — removes three files that should not have been committed.

Per project policy plans, repro playbooks, and ad-hoc diagnostic harnesses
are not project artifacts; they belong to the developer's local workspace.

## Removed

- `.repro/1257-sort-dropped-reference-attribute.md` — fix playbook. Also
  stale: described a fix path (filtering in `Attributes.getAttribute`) that
  was not the one ultimately adopted (decorator wrapping in `EntityDecorator`).
- `evita_test/.../MilagroCzReproductionTest.java` — manual diagnostics
  harness gated by `-Dmilagro.repro=true`, references a local production
  snapshot path, catches `Throwable` into a map without asserting (it would
  report success even on regressions). Not a real CI test.
- `evita_test/.../SortByDroppedReferenceAttributeTest.java` — its 3-product
  fixture does not cross the planner threshold that selects the
  `PrefetchedRecordsSorter` path, so it passed before the fix as well. A
  false-green regression test is worse than no regression test.

## Added

- `.gitignore` rule for `.repro/` so the directory cannot be tracked again,
  plus `.claude/scheduled_tasks.lock` which kept showing up as untracked.

The verification of #1257 stays in local developer notes and on the
production snapshot reproducer; it is not part of the committed sources.

Ref: #1257


novoj:
Closing without merging — would have triggered the release pipeline. Replacement PR uses `[skip ci]` in the commit message so no draft release is cut.

Issue #1332, opened 2026-07-28
perf: sort-attribute insert path — 7 bottlenecks from live WARM_UP profiling, plus the JMH coverage gap that hides two of them


novoj:
## Summary

A live async-profiler attach to an embedded evitaDB doing a WARM_UP bulk ingest surfaced seven distinct
bottlenecks on the **sort-attribute insert path**. `SortIndex.addRecordInternal` accounts for **57.5% of
busy-thread wall time** and **18.9% of allocation volume** during ingestion.

Two of the findings are algorithmic, four are pure allocation waste, one is a redundant tree descent. They are
independent and can land separately.

Critically, **the existing JMH suite structurally cannot observe the two algorithmic findings** — every sort-index
benchmark generates a value distribution in which the expensive branch is never taken. Closing that gap is a
prerequisite for measuring any of this, and is specified in detail below.

## How it was measured

- async-profiler 4.4, dynamic attach to a running JVM (no special launch flags), evitaDB `2026.2.0` embedded.
- Two 60 s windows: `-e alloc -o jfr` (**44.4 GB** of allocation sampled) and `-e wall -t -o jfr`.
- Wall shares are computed **within the single busy ingestion thread** (a merged wall profile across 114 threads is
  ~99% idle-futex noise and is meaningless — always filter on the thread's root frame).
- Allocation shares are contention-independent and trustworthy; wall shares are within-thread and fine, but they
  are **not** transactional-path or query-path numbers.

### Workload shape (this is the load-bearing detail)

The driver declares, per `offer` entity:

- **40 sortable `Integer` attributes** filled from `numberBetween(0, 1000)` — **1000 distinct values**, so each
  value-block holds `N/1000` records and grows linearly with the catalog.
- **5 sortable `OffsetDateTime` attributes** filled from `date().past(5, DAYS)` — effectively unique, block width 1.
- 30 faceted references, 5 non-indexed strings, 10M target entities.

Low-cardinality sortable attributes are **ordinary e-commerce**, not a synthetic edge case: rating 1-5, stock
level, discount band, brand id, priority, availability flag, category depth. The suite currently has no such
attribute anywhere (see "JMH coverage gap").

---

### 1. `computePreviousRecord` re-descends from the root once per binary-search probe

`SortIndexChanges.computePreviousRecord:228-267` binary-searches the value's block through
`sortIndex.sortedRecords.get(middle)` → `TransactionalUnorderedIntArray.get:335` →
[...]

Issue #1342, opened 2026-07-30
perf: collation cost on the sort-index write path (rankOf routing + per-transaction value-tree rebuild)


novoj:
Sorting a localized attribute routes every comparison through the JVM collator, which is roughly two orders of
magnitude more expensive than comparing two pre-computed collation keys. Profiling a 972 611-article localized
import showed that cost concentrated in one place, and measuring the fix revealed a second, larger defect on the
transactional write path. This issue covers both.

## Part 1 — `rankOf` routed internal nodes linearly (fixed)

`CumulativeWeightBPlusTree.rankOf` walked an internal node's separators **linearly** — up to 63 comparisons per
node — while `descend()`, five lines away in the same class, already binary-searched them in 6. On a localized
attribute every one of those comparisons may cost a full collation, and `rankOf` runs per record via
`SortIndexChanges.computePreviousRecord` as well as on the query path via `computeBlockStart`.

It owned **83 % of all collation-key computation** (11.84 % of process CPU, 17.40 % of allocation).

Fixed by extracting a shared `routeChildIndex` binary search used by *both* the routing and the rank path — so the
two can no longer disagree — and by adding `rankAndWeightOf`, which returns rank and weight from a single descent
as a packed long. `computePreviousRecord` previously made **three** root-to-leaf descents of the identical value.

**Measured** (quiet machine, arms compared at equal position, ≤1 % within-arm spread):

| | |
|---|---|
| Full CNC import, 972 611 articles | 416 s → **379 s** (1.098x) |
| Sort-index-dominated write path, 640k distinct values | **1.42x** (1.12x at 320k, 1.00x at 80k) |
| Comparator calls per lookup | 118.6 → **21.3** (5.58x fewer) |

The gain grows with the number of distinct indexed values, which is the mechanism's own prediction — the old
linear scan cost more the wider and deeper the tree got.

## Part 2 — collation-key cache retention (added, opt-in)

The caches legitimately fill during a bulk import and retain ~254 MB per locale to do it. Added
`ServerOptions.dropCollationKeysAfterSecondsOfInactivity` (CLOCK second chance; `0` = retain, the default).

Measured: releasing keys costs a bulk import **nothing** (379 s either way), but costs the **first write
transaction after a quiet spell** — 7.4 s → 12.9 s at 640k distinct values — while returning ~146 MB. That
penalty is a symptom of Part 3, not of the release, which is why it ships disabled. Note a single sweep reclaims
[...]

PR #1354, opened 2026-07-31
test: quantify the sort-attribute ingest benchmark's fixture with a control

### LINKED ISSUE #1335: test: sort-attribute ingest benchmark reports fixture allocation as ingest allocation
## Summary

`SortAttributeIngestBenchmark` (added in #1332) reports `gc.alloc.rate.norm` as its headline metric, but the figure includes the per-invocation fixture as well as the measured ingest. The **absolute** difference between two runs is still exact; the **percentage** is diluted and understates any improvement.

## Mechanism

`SortAttributeIngestBenchmarkState.setUp()` is `@Setup(Level.Invocation)` and per invocation:

- boots a full Evita instance,
- defines a 75-attribute / 30-reference schema,
- materialises 20 000 `InitialEntityBuilder` instances.

JMH's `GCProfiler` is an `InternalProfiler`: it snapshots allocation counters in `beforeIteration` / `afterIteration`, bracketing the **entire iteration including `Level.Invocation` fixtures**. So the reported bytes/op is `fixture + ingest`.

## Impact

For the A/B that motivated the benchmark, the fixture is byte-identical across both builds (seeded `Random` walked in primary-key order; `InitialEntityBuilder` is `evita_api` code untouched by the change), so:

- **the absolute delta is exact** — the fixture cancels in the subtraction;
- **the percentage is of the wrong denominator** — it has fixture in it, so the true ingest-only reduction is larger by an unquantified amount.

The #1332 results have already been restated in absolute GB/op terms, so nothing is currently misreported. This issue is about making the benchmark able to answer the percentage question at all.

The same mechanism affects `SortIndexTimingBenchmark.insertRecord`, whose `@Setup(Level.Invocation)` rebuilds 100 000–1 000 000 records against a measured batch of 10 000 — a 10:1 to 100:1 fixture-to-signal ratio on the allocation counter. Its JavaDoc has been corrected to state that only `ns/op` is trustworthy there (`SingleShotTime` times the benchmark method only, so the rebuild stays out of the wall clock). Its fixture genuinely cannot be coarsened — anything wider lets block widths drift away from the scenario definition — so documentation is the whole fix for that one.

## Proposed fix

1. **Add a control `@Benchmark`** with an empty body taking the same state. Whatever it reports *is* the fixture cost, so ingest allocation becomes `warmUpIngest - control`. The Evita boot genuinely cannot leave `Level.Invocation` — a catalog holding the previous invocation's entities would start with wide sort blocks and stop measuring a cold WARM_UP load — so subtraction is the only honest route.
2. **Memoise the entity batch** so it is built once per trial rather than once per invocation. The batch is deterministic and identical every invocation, so this changes nothing about what is ingested. (`InitialEntityBuilder.toMutation()` builds a fresh mutation list from immutable builder state on each call and neither consumes nor clears it, so re-use across invocations is safe.)
3. **Add `@Threads(1)`.** With `@State(Scope.Benchmark)` plus `Level.Invocation`, running under `-t N` would have N threads concurrently creating and closing the shared Evita instance.
4. **Restate the JavaDoc** to describe the control-subtracted protocol and name the JMH mechanism, so the caveat cannot be lost again.

## GO/NO-GO

> Run the control benchmark **before** changing anything else. If `control` is a small share of `warmUpIngest`, items 1 and 4 are enough and item 2 is not worth the churn. If `control` is the same order as `warmUpIngest`, do all four.

## Note

Changing the harness means the currently recorded figures are no longer reproducible against it — re-measure after the change and supersede the numbers in the #1332 measurement record rather than leaving both.

Ref: #1332

novoj:
Closes #1335

`SortAttributeIngestBenchmark` reports `gc.alloc.rate.norm` as its headline metric, but JMH's `GCProfiler` is an `InternalProfiler`: it snapshots the allocation counters in `beforeIteration` / `afterIteration`, bracketing the entire iteration **including** the `@Setup(Level.Invocation)` fixture and its teardown. The reported bytes/op was therefore `fixture + ingest` — absolute A/B deltas were exact, but percentages carried fixture allocation in the denominator.

## What changed

- **`fixtureControl`** — a second `@Benchmark` with an empty body taking the same state. JMH runs both `Level.Invocation` fixtures around it regardless, so what it reports *is* the fixture. `warmUpIngest − fixtureControl` is the ingest's own allocation.
- **`@Threads(1)`** — documented as declarative only, since JMH's `-t` still overrides an annotation.
- **JavaDoc** — the subtraction protocol, its three conditions, and reproduction instructions.

## GO/NO-GO gate

The issue prescribed running the control *before* changing anything else. Measured at `entityCount = 20 000` (JDK 21.0.11, `-Xmx8g`, `-wi 1 -i 3`, one run so both halves of each subtraction share machine state):

| `distinctValues` | `fixtureControl` | `warmUpIngest` | control share | ingest-only |
|---|---|---|---|---|
| 1000 | 1.268 GB/op | 94.224 GB/op | 1.35 % | 92.956 GB/op |
| 20 | 1.257 GB/op | 76.026 GB/op | 1.65 % | 74.769 GB/op |

The fixture is a small share, so **item 2 of the issue (memoising the entity batch) was declined**: the subtraction already removes all of the fixture, exactly, so hoisting would buy no accuracy — only churn and an uneven first iteration. That holds even though the batch turns out to be ~96 % of the fixture itself.

## Two conditions the issue didn't anticipate

**Subtract per `distinctValues`, never pooled.** `setAttribute` boxes every low-cardinality `int` and `Integer.valueOf` caches `−128..127`, so at `distinctValues = 20` all values are cache hits and allocate nothing, while at 1000 only 12.8 % are. The measured gap between the two controls is 11.165 MB against 11.16 MB predicted from the 697 600 boxes that escape the cache. A pooled control would over-subtract one arm and under-subtract the other.

**The remainder carries a teardown asymmetry.** `closeEvita()` is inside the profiler bracket for both methods, but the control closes an *empty* catalog while `warmUpIngest` closes one holding `entityCount` entities. The flush the ingest provokes stays in the remainder — the right place for it, but the remainder is not the `upsertEntity` loop in isolation.

## Reproducibility finding

The #1332 figures were taken with `-p entityCount=5000`, **not** the committed `@Param` default of 20 000 — the block widths quoted beside them (~5 and ~250) only follow from 5 000. At that override this harness reproduces them to within 1.2 %, which confirms `warmUpIngest` is unchanged by this PR and the recorded absolutes remain valid. PR #1336's body has been updated with the restated ingest-only percentages: **12.92 %** and **15.13 %** (previously 12.76 % and 14.93 % raw).

## Not changed

`SortIndexTimingBenchmark` is affected by the same `GCProfiler` mechanism at a 10:1–100:1 fixture-to-signal ratio, but its fixture cannot be coarsened without letting block widths drift off the scenario definition, so documentation is the whole fix there — and `75cf26b50` already applied it. Its JavaDoc states that only `ns/op` is trustworthy; verified, left alone.

### DIFF
--- a/evita_test/evita_performance_tests/src/main/java/io/evitadb/performance/sortattribute/SortAttributeIngestBenchmark.java
+++ b/evita_test/evita_performance_tests/src/main/java/io/evitadb/performance/sortattribute/SortAttributeIngestBenchmark.java
@@ -33,6 +33,7 @@ import org.openjdk.jmh.annotations.Fork;
 import org.openjdk.jmh.annotations.Measurement;
 import org.openjdk.jmh.annotations.Mode;
 import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Threads;
 import org.openjdk.jmh.annotations.Warmup;
 import org.openjdk.jmh.infra.Blackhole;
 
@@ -60,11 +61,12 @@ import java.util.concurrent.TimeUnit;
  * measured cross-jar against the base engine. It is {@link Mode#SingleShotTime} so entities per second is
  * `entityCount / score`, but at the iteration counts it can afford the wall-clock number is far too noisy to trust
  * (this issue saw ±258 % on `ns/op` against ±0.75 % on allocation in the same run), so **read `gc.alloc.rate.norm`,
- * not the score**, and run with `-prof gc` - reading that figure the way {@link SortAttributeIngestBenchmarkState}
- * prescribes, because the per-invocation fixture's allocation is counted inside it. Note that the CPU-only changes
- * are invisible here: the retained-leaf block search alters no allocation, and resolving the thread's transaction
- * once per positional read removes a single duplicate `ThreadLocal` read that no end-to-end wall-clock measurement
- * can separate from noise - it is kept on strictly-fewer-operations grounds, not on a measured end-to-end delta.
+ * not the score**, and run with `-prof gc` - subtracting {@link #fixtureControl} from it exactly as
+ * {@link SortAttributeIngestBenchmarkState} prescribes, because the per-invocation fixture's allocation is counted
+ * inside the raw figure. Note that the CPU-only changes are invisible here: the retained-leaf block search alters no
+ * allocation, and resolving the thread's transaction once per positional read removes a single duplicate
+ * `ThreadLocal` read that no end-to-end wall-clock measurement can separate from noise - it is kept on
+ * strictly-fewer-operations grounds, not on a measured end-to-end delta.
  *
  * Heap is sized explicitly in the fork arguments on purpose. The profiled process ran without `-Xmx`, so JVM
  * ergonomics handed it 23.4 GiB and it sat at 92 % old-gen occupancy with the young generation squeezed to 320 MB;
@@ -88,6 +90,10 @@ import java.util.concurrent.TimeUnit;
 		"-Xmx8g"
 	}
 )
+// the state is Scope.Benchmark and its fixture is Level.Invocation, so under `-t N` all N threads would create and
+// close the one shared Evita instance concurrently. JMH's command line still wins over this annotation - `-t N`
+// overrides it - so treat it as the declared intent, not as an enforced guard: never pass `-t` to this benchmark.
+@Threads(1)
 public class SortAttributeIngestBenchmark implements EvitaCatalogSetup {
 
 	/**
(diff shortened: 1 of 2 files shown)


novoj:
Superseded by `950c803f9` on `dev`, which landed a fix for #1335 (bundled into the #1342 performance commit) before this branch was opened. That version also implements the batch memoisation this PR declined, so it is strictly ahead. Closing as duplicate.

One finding from this branch is not carried by `dev` and is being opened as a small follow-up: the #1332 end-to-end figures were measured with `-p entityCount=5000`, not the committed `@Param` default of `20000` — the block widths quoted beside them (~5 and ~250) only follow from 5 000. Without that recorded, those numbers cannot be reproduced.

PR #1336's body has already been updated with the control-subtracted percentages (12.92 % and 15.13 %), computed against the original harness those figures were measured on.

Issue #1392, opened 2026-08-05
Reference-key equality mechanism: replace the wrapper-based non-uniform internal-PK handling


novoj:
**Timing:** 2026.3 or later — this is a design fork, not a refactor.

## Context

`ReferenceKey` carries an `internalPrimaryKey` in three bands: `0` unknown (client boundary, before
the server sees the reference), negative non-persistent (assigned immediately on server-side
arrival, `ReferencesStoragePart.assignMissingIdsAndSort`), positive persistent. The internal PK
exists to let duplicate references (same `referenceName` + `primaryKey`) coexist.

Because `ReferenceKey.equals` is deliberately lenient — it compares `internalPrimaryKey` only when
**both sides are positive** (`ReferenceKey.java:137`), and `hashCode` covers only
name + primaryKey — the structures that must *distinguish* duplicates while internal PKs are still
negative/mixed cannot key on `ReferenceKey` directly. Two wrapper types exist solely to supply a
stricter equality:

- **`ComparableReferenceKey`** (`evita_api`, public): `equals` delegates to `FULL_COMPARATOR`,
  which compares internal PKs whenever *neither* side is unknown (`0`) — so distinct negative
  duplicates stay distinct; `hashCode` includes `internalPrimaryKey` unconditionally.
- **`FullyComparableReferenceKey`** (`ExistingReferencesBuilder.java:2160`, private): strict record
  equality on all three fields, asserting a known internal PK at construction.

### These wrappers are functionally load-bearing — verified per site

- **`ReferencesStoragePart.assignMissingIdsAndSort` result map (`:236`)** — maps *pre-assignment*
  keys (negative/unknown) to newly assigned keys. Duplicate references with distinct negative PKs
  would collide under lenient `equals` (both non-positive ⇒ equal, same hash) and the second `put`
  would silently overwrite the first assignment.
- **`ReferencesStoragePart.referenceKeysForReassignment` (`:113`, `:228`)** — membership must target
  the *specific* duplicate being reassigned, not any reference sharing the business key.
- **`ContainerizedLocalMutationExecutor.assignedPrimaryKeys` (`:3621`, `:3674`, `:3780`)** — reverse
  translation of original to assigned key during mutation propagation; same collision hazard as the
  first site.
- **`ReferenceAttributeValueProvider.referenceIndex` (`:155`–`:166`)** — per-reference contract
  lookup; duplicates with distinct negative PKs must remain distinct entries (lookup keys asserted
[...]

PR #1393, opened 2026-08-05
perf: rewrite verifyReferenceCardinalities as an allocation-free run-length scan

### LINKED ISSUE #1391: Rewrite verifyReferenceCardinalities as an allocation-free run-length scan
**Delivery:** hotfix on `release_2026-2`

## Context

`ContainerizedLocalMutationExecutor.verifyReferenceCardinalities`
(`ContainerizedLocalMutationExecutor.java:3044`) is the **#1 allocator (~21%) and #2 CPU consumer
(~11%)** in the Senesi WARM_UP profile (2026.2.2, see
`specifications/write-path-optimizations/README.md` §2–§3), and it builds no index and writes no
byte. Every `ObjectIntHashMap` sample in the profile (12.77% inclusive, of which `rehash` alone is
4.95%) is attributable to this one method — the profile's apparent "hppc bottleneck" is this method
seen from below.

Per entity it allocates two `ObjectIntHashMap`s (one per reference *name*, one per `ReferenceKey`;
the latter unsized, so a reference-heavy entity pays ~6 growth/rehash cycles).

## Why a linear scan is possible

`ReferencesStoragePart.getReferences()` returns an array sorted by
`ReferenceContract.FULL_COMPARATOR` (`referenceName` → `primaryKey` → `internalPrimaryKey`). The
invariant is assert-enforced (`ReferencesStoragePart:216`) and re-established with `Arrays.sort`
after every modification (`:245`, `:772`). The method already leans on it (reflected-reference
skip, `referenceSchema` memo assume contiguous same-name runs).

## What to do

Rewrite the method as a single pass over the sorted array:

- **Cardinality per name:** run-length count over each contiguous same-name run; check the limit at
  the run boundary. No map.
- **Duplicate detection:** within a run, equal keys are adjacent — compare each element to its
  predecessor with `ReferenceKey.equals` (NOT `primaryKey ==`). No map.
- **Min-cardinality (required but absent):** the only cross-run state; a small reused structure of
  seen names suffices (~20 names). Beware: `EntitySchema.getReferences()` is a `LinkedHashMap` in
  *insertion* order, not sorted — a merge-join against the runs is wrong without sorting it first.

**Preserve current semantics exactly — do not redesign key equality here.** `ReferenceKey.equals`
compares `internalPrimaryKey` only when both sides are positive; the write path deliberately keeps
keys in *uniform knownness form* within any one structure (unknown internal PKs exist only at the
external boundary; the server assigns negative, then positive PKs immediately on arrival). The
rewrite must reproduce today's grouping behaviour under that invariant, and must document the
invariant in a comment at the site. Any change to the equality mechanism itself is out of scope —
tracked in #1392.

## Rejected alternative — recorded so it is not re-proposed

**Skip validation entirely on "trusted" bulk loads: rejected.** Nothing on this path can be
trusted; and once the rewrite lands, the validation costs nearly nothing, so a skip flag would buy
a class of only-when-off bugs for no measurable win.

## Fallback

If the rewrite is deferred for any reason, the minimal fix is a one-line size hint at
`ContainerizedLocalMutationExecutor.java:3077` (`new ObjectIntHashMap<>(cntReferences.length)`),
which removes the rehash cycles but not the hashing, allocation, or iteration.

## Verification

Re-run the Senesi WARM_UP benchmark: `ObjectIntHashMap` frames should disappear from the profile
entirely (they have no other owner). Full test suite green; existing cardinality-violation tests
must pass unchanged.

## Risk

Self-contained rewrite of one private method with an exact behavioural contract; the sortedness it
depends on is assert-enforced.

novoj:
Closes #1391

## What

`ContainerizedLocalMutationExecutor.verifyReferenceCardinalities` was the **#1 allocator (~21%) and #2 CPU consumer (~11%)** of the Senesi WARM_UP profile while building no index and writing no byte. Per entity it allocated two `ObjectIntHashMap`s — one keyed by reference *name*, one by `ReferenceKey` (the latter unsized, so a reference-heavy entity paid ~6 growth/rehash cycles). Every `ObjectIntHashMap` sample in that profile belonged to this one method.

Both maps are unnecessary: `ReferencesStoragePart#getReferences()` is sorted by `ReferenceContract.FULL_COMPARATOR`, so references sharing a name form a contiguous run and equal keys are adjacent within it. The method is now a single run-length pass that **allocates nothing unless a violation is actually found**:

- **per-name cardinality** — the run length; the upper limit is checked at the run boundary;
- **duplicates** — the examined key is compared against the *head of the current cluster* of equal keys. Anchoring on the cluster head rather than on the immediate predecessor is deliberate: it is what reproduces the "first key wins as the canonical one" grouping of the hash map being replaced, and it stays faithful even for a run mixing known and unknown internal PKs, where predecessor-comparison would not;
- **min cardinality** — the only cross-run state; presence is resolved by a lower-bound binary search on the reference name and evaluated only for mandatory references, which are rare.

The three now-unused `hppc` imports are gone from the file.

## Semantics

Preserved exactly, per the issue. That includes the **currently unreachable upper-limit branch** (`getMax() <= 0`): no `Cardinality` constant declares a maximum lower than one, and the effective upper-limit check lives in `InitialReferencesBuilder` / `ExistingReferencesBuilder` (`getMax() <= 1`). It is kept verbatim with a comment naming both facts rather than silently deleted or tightened — tightening it would be a behavioural change out of scope here.

The `ReferenceKey.equals` conditional-equality hazard is documented at the site, not redesigned (that remains #1392): the method runs from `verifyConsistency()`, i.e. after `finishLocalMutationExecutionPhase()` has assigned terminal positive internal PKs, so every key it sees is in uniform *known* form.

## One deliberate behavioural delta

The old map-based count was **order-insensitive** — `putOrAdd` grouped by name regardless of contiguity, so latent disorder in a container would still have produced correct counts. The new scan is order-sensitive, and the only existing sortedness assert (`ReferencesStoragePart#assignMissingIdsAndSort`) sits behind `if (this.unassignedPrimaryKeys)`, reached solely when a *new* reference is inserted — an attribute-only change, a reference removal or a group change never reaches it.

The scan therefore verifies the ordering itself, so **a disordered container now fails loudly with `GenericEvitaInternalError` instead of being validated against a map that could not notice it**. Cost of that tripwire: one `Integer.compare` plus a never-taken branch per reference, and one `String.compareTo` per name run — it deliberately replays only the *numeric tail* of `FULL_COMPARATOR`, since inside a run the names are already known equal and re-comparing them would double the loop's string work. A cross-reference was added at `ReferencesStoragePart.references` so that anyone changing sort maintenance sees the new consumer.

## Verification

- `ReferenceIndexingTest` — **15/15 green**; 10 in the cardinality group: 6 pre-existing **unchanged**, 4 new. The new cases cover the run *boundary*, which nothing exercised before (every existing test used a single reference name): multi-name happy path, mandatory-absent-while-other-names-present, all missing mandatory references reported together, and mandatory reference removed on an existing entity (the dropped-run path).
- Full `unitAndFunctional` — **20838 run, 2 failures, both unrelated**: `ExportS3ServiceTest` (no Docker environment available) and `EvitaTest.shouldKillInactiveSessionsAutomatically`, which passes when re-run in isolation — timing-flaky under suite load.
- **The Senesi WARM_UP re-run named in the issue is not executable**: ADR `2026-07-27-write-path-performance-tuning` records that the catalog snapshot and WAL slice every measurement was taken against were deleted. Re-measuring needs a fresh production export.

## Known coverage limit

The duplicate-cluster logic has no direct test and cannot get one at this level: under the uniform-knownness invariant, identical triples cannot coexist in a strictly sorted array, so the branch is unreachable by construction. It is preserved as the defensive net it already was.

### DIFF
--- a/evita_engine/src/main/java/io/evitadb/index/mutation/storagePart/ContainerizedLocalMutationExecutor.java
+++ b/evita_engine/src/main/java/io/evitadb/index/mutation/storagePart/ContainerizedLocalMutationExecutor.java
@@ -24,9 +24,6 @@
 package io.evitadb.index.mutation.storagePart;
 
 import com.carrotsearch.hppc.IntHashSet;
-import com.carrotsearch.hppc.ObjectIntHashMap;
-import com.carrotsearch.hppc.ObjectIntMap;
-import com.carrotsearch.hppc.cursors.ObjectIntCursor;
 import io.evitadb.api.exception.EntityMissingException;
 import io.evitadb.api.exception.InvalidMutationException;
 import io.evitadb.api.exception.MandatoryAssociatedDataNotProvidedException;
@@ -3036,6 +3033,28 @@ public final class ContainerizedLocalMutationExecutor
 	 * Ensures that each reference complies with its respective cardinality requirements. If any violations are found,
 	 * a {@link ReferenceCardinalityViolatedException} is thrown with details of the violations.
 	 *
+	 * The verification is a single run-length scan of the reference array that allocates nothing unless a violation
+	 * is actually found. A plain scan is sufficient because {@link ReferencesStoragePart#getReferences()} is always
+	 * sorted by {@link ReferenceContract#FULL_COMPARATOR} - i.e. by reference name, then referenced primary key and
+	 * finally by internal primary key. The invariant is asserted in
+	 * {@link ReferencesStoragePart#assignMissingIdsAndSort()} and re-established by re-sorting the array after each
+	 * modification of the container. It implies that:
+	 *
+	 * - references sharing the same name form a single contiguous run, so the number of references of a particular
+	 *   name is a mere length of that run and needs no auxiliary map,
+	 * - references sharing the same key are adjacent within such run, so duplicates are recognized by comparing the
+	 *   examined key with the head of the currently examined cluster of equal keys.
+	 *
+	 * **Reference key equality invariant:** {@link ReferenceKey#equals(Object)} takes the internal primary key into
+	 * account only when it is known (positive) on both compared sides, and is therefore not transitive for a mixture
+	 * of keys with known and unknown internal primary keys. This method never observes such a mixture - it is called
+	 * from {@link #verifyConsistency()}, which runs only after {@link #finishLocalMutationExecutionPhase()} assigned
+	 * terminal (positive) internal primary keys to all the references of the container, so all the keys examined here
+	 * are in uniform - known - form and the equality degenerates to a plain comparison of all three key components.
+	 * The examined key is nevertheless compared against the cluster head and not against its immediate predecessor,
+	 * since that is what reproduces the "first key wins as the canonical one" grouping of the hash map this scan
+	 * replaced, should the uniformity invariant ever weaken.
+	 *
 	 * @param referencesStorageContainer An optional container holding the current state of references associated
 	 *                                   with an entity. Can be null if there are no references to validate.
 	 * @throws ReferenceCardinalityViolatedException If any reference violates its defined cardinality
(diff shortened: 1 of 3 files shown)


novoj:
Superseded by #1395, which combines this with #1394 into a single branch off `dev` with the `ContainerizedLocalMutationExecutor` interaction already resolved. Closing unmerged — the commit is carried over verbatim.

PR #1394, opened 2026-08-05
perf: cut schema-resolution cost and schema-derived key allocation on the write path

### LINKED ISSUE #1390: Schema-handling write-path optimizations: resolution cost + schema-derived key allocations
**Delivery:** hotfix on `release_2026-2`

## Context

Two clusters of the Senesi WARM_UP profile (2026.2.2, see
`specifications/write-path-optimizations/README.md` §1 and §4) have the same root: schema facts and
schema-derived keys are re-resolved or re-allocated per mutation on a path where the schema cannot
change.

**CPU — 18.7% of write path** spent resolving facts from immutable schema objects:

| frame | CPU share |
|---|---|
| `EntitySchema.getReference` | 15.36% |
| `ReferenceSchema.getIndexedComponents` | 1.91% |
| `ReferenceSchema.isFacetedInScope` | 1.43% |

**Allocation — 18.24% of write path** in `AttributeKey` — short-lived `(attributeName, locale)`
records created per reference per entity purely to be compared or looked up, then discarded.

There are 36 schema-resolution call sites in `io.evitadb.index.mutation`, 15 of them in
`ContainerizedLocalMutationExecutor` alone.

**Audit result (2026-08-05, dev = v2026.2.2 for all schema DTOs — verified by diff):** the callee
bodies of the two smaller CPU frames are already near-minimal — `isFacetedInScope` is a `contains`
on an `EnumSet`-backed set (`ReferenceSchema.java:1008`, built via `ArrayUtils.toEnumSet` at
`:639`), `getIndexedComponents` a plain map `get` (`:996`). Their profile share is therefore
**call frequency**, not body cost. The one genuinely fat accessor is `getReferenceOrThrowException`.

## Part 1 — allocation-free "or throw" accessors

`EntitySchema.getReferenceOrThrowException` (`EntitySchema.java:933`) allocates two `Optional`s and
a capturing lambda per call; replace with plain map `get` + null-check + throw.
`getAssociatedDataOrThrowException` (`:884`) has the identical pattern — fix it the same way.

## Part 2 — hoist schema resolution out of the per-mutation loop

Since the callee bodies are already cheap, the 15.36% is volume: the 36 call sites re-resolve the
same reference schema per mutation/reference. References are processed in name-sorted runs, so the
schema (and derived facts like `getIndexedComponents(scope)` / `isFacetedInScope(scope)`) can be
resolved **once per same-name run or once per entity** and carried in a local. This — not callee
memoization — is what removes the call volume.

## Part 3 — memoize the accessors that genuinely re-derive per call

`ReferenceSchema.getIndexedInScopes()` builds a stream + new `Set` on every call
(`ReferenceSchema.java:974`–`:980`). `EntitySchema` already has the memoization precedent
(`memoizedLocalized`, `EntitySchema.java:789`). Audit the accessors reached from the write path for
the same shape.

## Part 4 — canonical `AttributeKey` cache on `AttributeSchema`

The schema-derived allocation sites build an `AttributeKey` from `attributeSchema.getName()` plus a
locale from a small entity-locale set, use it for one lookup/comparison, and discard it. The domain
is tiny and schema-bounded — cache canonical instances on the schema: a precomputed locale-agnostic
`AttributeKey` plus a small lazy `Locale → AttributeKey` map, living as long as the schema. Usage
is equality-based everywhere and nothing relies on instance identity, so this is behaviour-neutral.

Known hot sites: `ReferenceBlock.java:237/248/273/292`,
`ContainerizedLocalMutationExecutor.java:259/265/1298/2129/2136`, `AttributeIndexMutator.java:896`,
`ReferenceIndexMutator.java:1803`.

**Honesty note:** the profile does not attribute the 18.24% between schema-derived and
mutation-embedded families — measure per-site (alloc profile) before/after to confirm the win.

## Part 5 — allocation-free `AttributeKey.compareTo`

`AttributesContract.AttributeKey.compareTo` (`AttributesContract.java:352`) allocates a capturing
lambda per comparison via `ComparatorUtils.compareLocale(..., () -> ...)`. Rewrite without the
supplier. This fires inside every sorted structure keyed by `AttributeKey`.

## Part 6 — schema-family accessor sweep (audit-backed residue)

The audit of the whole `schema/dto` family found per-call accessors largely already clean — plain
map/set lookups, scope containers `EnumSet`-backed, streams confined to constructors, `toString`
and one-time conversion helpers. The sweep targets the concrete residue, not a blanket rewrite:

- `ReferenceSchema.isIndexedInScope` (`:968`–`:969`) — double lookup (`containsKey` + `get`);
  collapse to one `get`.
- `AttributeSchema.getUniquenessType` (`:474`) — `ofNullable().orElse()` per call; replace with a
  plain null-check.
- **Internal `@Nullable` accessors for hot Optional getters.** `getAttribute` / `getReference` /
  `getAssociatedData` return `Optional.ofNullable(map.get(name))` — one allocation per call on
  paths that call them millions of times. The public `Optional` contract stays; hot internal
  callers in `io.evitadb.index.mutation` get direct null-returning variants, per the project rule
  that `Optional` belongs at the read boundary.
- Verified clean, no action: `AttributeSchema` scope predicates (`isFilterableInScope` etc.),
  `AssociatedDataSchema` (plain holder), `ReflectedReferenceSchema` overrides (constant-message
  asserts + super delegation), lazily-memoized `inheritedAttributes`.

## Explicitly out of scope

- **Name canonicalization** (the identity-miss on every string-keyed `HashMap` lookup, 21.36%
  `getNode`) — needs design exploration and a measured ceiling first; see
  `specifications/write-path-optimizations/exploration-name-canonicalization.md`. That also covers
  `AttributeKey` instances embedded in deserialized mutations (`UpsertAttributeMutation` etc.,
  fresh per gRPC message).
- **`ComparableReferenceKey` removal** — functionally load-bearing for the non-uniform internal-PK
  mechanism; inseparable from the equality redesign tracked in #1392. Must not be
  attempted here.

## Verification

One Senesi WARM_UP re-run covers both clusters: the three CPU frames above should drop
substantially, and schema-derived `AttributeKey` allocation should shrink with the residual
attributed to the mutation-embedded family. Full test suite green.

## Risk

None to on-disk format or query semantics — accessor refactors and additive caches on immutable
objects, plus local hoisting in the mutation executor.

novoj:
Closes #1390

Delivers all six parts of the issue. The Senesi WARM_UP profile of `2026.2.2` attributes 18.7% of
write-path CPU to resolving facts from immutable schema objects and 18.24% of write-path allocation
to short-lived `AttributeKey` records. Both are **call volume**, not body cost — the accessors are
already plain map lookups, and the keys are re-derived from a schema that cannot change while the
mutation is being applied.

Decision record: `documentation/adr/2026-08-05-schema-handling-write-path-optimizations.md`.

## What changed

| Part | Change |
|---|---|
| 1 | `EntitySchema.getReferenceOrThrowException` / `getAssociatedDataOrThrowException` → map get + null-check + throw, dropping two `Optional`s and a capturing lambda per call |
| 2 | `indexAllReferences` / `unindexReferences` / `propagateOrphanedReferenceAttributeMutations` resolve the reference schema **and its derived facts** once per name-sorted run; `updateReferences` resolves once instead of up to three times per mutation |
| 3 | `ReferenceSchema.getIndexedInScopes()` precomputed in the constructor as an unmodifiable `EnumSet` |
| 4 | `AttributeSchema` owns canonical `AttributeKey` instances (one locale-agnostic + a lazy `Locale`-keyed map); 8 hot sites rewired |
| 5 | `AttributeKey.compareTo` inlines the `ComparatorUtils.compareLocale` semantics instead of allocating a capturing `IntSupplier` per comparison |
| 6 | `isIndexedInScope` single lookup; `getUniquenessType` null-check; `EntitySchema.getAttributeOrNull` for `updateAttributes` |

## Safety

- **No on-disk format change, no `serialVersionUID` bump.** Schema DTOs are written by hand-written
  serializers that enumerate fields explicitly and are versioned through `SerialVersionBasedSerializer`,
  so the derived cache fields are invisible to persistence.
- **Sharing `AttributeKey` instances is behaviour-neutral.** Verified: no `IdentityHashMap` and no `==`
  anywhere keys on `AttributeKey`, and Kryo runs with `setReferences(false)` (`KryoFactory:120`), so the
  serialized graph of the WAL-bound mutations is unchanged.
- **The hoisted derived facts keep their original guards.** `indexedForEntityComponent` is computed only
  under `indexedForFiltering`, `faceted` only under `indexedForEntityComponent` — because
  `ReflectedReferenceSchema.isFacetedInScope` asserts on reflected-reference availability and must not
  start firing where the old control flow never called it.
- Both cache fields on `AttributeSchema` are `@EqualsAndHashCode.Exclude`; without that the lazily
  populated map would make two otherwise-equal schemas compare unequal depending on which locales had
  been touched.

## Verification

- `AttributeKeyTest` (new, 7 tests) pins the full `AttributeKey` ordering contract — all four
  null/non-null locale combinations, locale-string ordering, the tie-break onto the attribute name, and
  an exhaustive cross-check of every (locale, name) pair against `ComparatorUtils.compareLocale`. It was
  written and **run green against the previous implementation first**, so it pins the historical
  semantics rather than the rewrite's.
- `evita_engine` and all upstream modules compile clean.
- Full `evita_functional_tests` suite: **still running locally at the time this PR was opened.** CI is
  the authority here; I will confirm once it reports.

## Not done, deliberately

- **No performance number is claimed.** The Senesi WARM_UP re-run named in the issue's verification
  section has not been executed — it is a multi-hour two-process job in a separate worktree. The
  percentages above are the measured cost of the *previous* code, from the profile that motivated the
  issue; the expected wins are argued from reading each site.
- **Part 3's target is not actually on the write path.** `ReferenceSchema.getIndexedInScopes()` has no
  caller in `io.evitadb.index.mutation` — its only non-test callers are `ClassSchemaAnalyzer` and
  schema-mutation code. It is memoized as the issue asked, but it is a tidy-up, not a win. Recorded in
  the ADR so the next reader does not repeat the audit.
- **Two `AttributeKey` sites left alone** — `ReferenceIndexMutator.readReferenceAttributeValue` and
  `ContainerizedLocalMutationExecutor.readReferencedEntityAttribute`. Both are keyed by a bare attribute
  *name* from a histogram descriptor pointing at a **referenced entity's** attribute, so the owning
  schema is not in hand and may live in another collection.
- **The two `*AttributeAndCompoundSchemaProvider` classes** still resolve through the `Optional`
  accessors; narrowing their contract-typed fields to the DTOs would mean widening ~10
  `ReferenceSchemaContract` parameters through `ReferenceIndexMutator`.
- **Out of scope per the issue, recorded with reasons in the ADR:** reference-name canonicalization
  (needs a design exploration and a measured ceiling) and `ComparableReferenceKey` removal (load-bearing
  for the non-uniform internal-PK mechanism, inseparable from #1392).
- `specifications/write-path-optimizations/` lives on `dev`, not on this release branch, so the plan
  folder cannot be retired here.

### DIFF
--- a/documentation/adr/2026-07-27-write-path-performance-tuning/README.md
+++ b/documentation/adr/2026-07-27-write-path-performance-tuning/README.md
@@ -1,7 +1,7 @@
 ---
 title: Cut commit-merge latency and write-path allocation by pruning the trunk merge, not inverting it
 date: 2026-07-27
-updated: 2026-07-31 21:05
+updated: 2026-08-05 16:00
 status: accepted
 kind: optimization
 issues: [760]
@@ -9,7 +9,7 @@ prs: [1317, 1298]
 areas: [evita_engine/core/transaction, evita_engine/index, evita_engine/core/buffer, evita_engine/index/bPlusTree]
 supersedes: []
 superseded-by: []
-relates: [2026-07-10-more-optimized-data-structures]
+relates: [2026-07-10-more-optimized-data-structures, 2026-08-05-schema-handling-write-path-optimizations]
 ---
 
 # Write-path performance tuning — commit-merge latency and allocation
--- a/documentation/adr/2026-08-01-bplustree-cursor-free-insert-path.md
+++ b/documentation/adr/2026-08-01-bplustree-cursor-free-insert-path.md
@@ -1,7 +1,7 @@
 ---
 title: Answer the B+ tree insert-boundary asserts from the descent instead of a captured cursor path
 date: 2026-08-01
-updated: 2026-08-01 10:45
+updated: 2026-08-05 16:00
 status: accepted
 kind: optimization
 issues: [1333]
@@ -9,7 +9,7 @@ prs: [1356]
 areas: [evita_engine/index/bPlusTree]
 supersedes: []
 superseded-by: []
-relates: [2026-07-10-more-optimized-data-structures, 2026-07-31-bulk-ingest-write-path]
+relates: [2026-07-10-more-optimized-data-structures, 2026-07-31-bulk-ingest-write-path, 2026-08-05-schema-handling-write-path-optimizations]
 ---
 
 # Free the B+ tree insert and read paths from the cursor path
--- a/documentation/adr/2026-08-05-schema-handling-write-path-optimizations.md
+++ b/documentation/adr/2026-08-05-schema-handling-write-path-optimizations.md
@@ -1,0 +1,208 @@
+---
+title: Share schema-derived attribute keys and resolve reference schemas once per run instead of per mutation
+date: 2026-08-05
+updated: 2026-08-05 16:00
+status: accepted
+kind: optimization
+issues: [1390]
+prs: []
+areas: [evita_api/requestResponse/schema/dto, evita_api/requestResponse/data, evita_engine/index/mutation]
+supersedes: []
+superseded-by: []
+relates: [2026-07-27-write-path-performance-tuning, 2026-08-01-bplustree-cursor-free-insert-path]
+---
+
+# Share schema-derived attribute keys and resolve reference schemas once per run instead of per mutation
+
+The Senesi WARM_UP profile of `2026.2.2` attributes 18.7% of write-path CPU to resolving facts from
+immutable schema objects and 18.24% of write-path allocation to `AttributeKey` instances. Both are
+volume, not body cost: the accessors are already plain map lookups, and the keys are re-derived from
+a schema that cannot change while the mutation is being applied. This record covers the resulting
+work — allocation-free "or throw" accessors, canonical `AttributeKey` instances owned by
+`AttributeSchema`, an allocation-free `AttributeKey.compareTo`, and per-run hoisting of reference
+schema resolution in the mutation executors — and, more importantly, the two adjacent options that
+were deliberately **not** taken.
+
+## Why
+
+Bulk ingest applies millions of local mutations against one immutable entity schema. Two families of
+cost fall out of that:
+
+- **Resolution volume.** `EntitySchema.getReference` alone is 15.36% of write-path CPU, with
+  `ReferenceSchema.getIndexedComponents` (1.91%) and `isFacetedInScope` (1.43%) behind it. An audit
+  of the callee bodies found them already near-minimal — `isFacetedInScope` is a `contains` on an
+  `EnumSet`-backed set, `getIndexedComponents` a plain map `get`. The share is therefore how often
+  they are called, and the one genuinely fat accessor was `getReferenceOrThrowException`, which
+  allocated two `Optional`s and a capturing lambda per call to express "get or throw".
+- **Key churn.** Schema-derived `AttributeKey` records are built per attribute per reference per
+  entity purely to be compared or looked up, then discarded. The domain is tiny and
+  schema-bounded — one key per (attribute, locale) pair — while the allocation count scales with the
+  number of entities ingested.
+
+The constraint that makes this non-obvious is that the schema is **not** pinned for the duration of
+an entity's mutation batch. `ContainerizedLocalMutationExecutor` reads it through
+`this.schemaAccessor.get()`, a supplier re-read at every use site, because schema evolution can add
+a reference or an attribute part-way through applying a batch. Any caching strategy has to survive
+that.
+
+### Previous state
+
+`getReferenceOrThrowException` was `getReference(name).map(cast).orElseThrow(() -> new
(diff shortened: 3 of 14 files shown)


novoj:
Superseded by #1395, which combines this with #1393 into a single branch off `dev` with the `ContainerizedLocalMutationExecutor` interaction already resolved. Closing unmerged — the commit is carried over, plus an ADR correction: the spec's §2/§3 are no longer "unclaimed", they ship as the sibling commit.

PR #1417, opened 2026-08-12
Bind catalogs to opaque folder tokens, and make rename and replace a pointer swap

### LINKED ISSUE #649: Cannot replace catalog through gRPC
Seems like there is bug (probably specific to Windows machine) where it cannot rename directory when replacing the catalog.

Image

novoj:
Closes #649

`replaceCatalog(A, B)` failed intermittently over gRPC, and when it failed it left the operator to infer the state of two catalogs from whatever was on disk — the contract said as much: *"the state of `catalogNameToBeReplacedWith` is unknown and should be treated as damaged"*.

Windows made it frequent but was not the cause. The operation was a five-step filesystem mutation with no durable record of how far it had got: rename every file inside the source folder onto the target's name, move the target folder aside, move the source folder onto the target's path, delete the folder moved aside — and undo all of it on any failure. A reader holding a directory open makes step three fail, the compensating rename can fail too, and then there is no correct state to return to.

## What changed

A catalog's directory used to *be* its name — `storageDirectory.resolve(catalogName)`, in some thirty places. A catalog is now bound to an opaque `CatalogFolderId` recorded in the engine state; the folder's name is cosmetic, and **renaming or replacing a catalog is a single engine-state commit that moves nothing on disk**. A replace points the target's name at the source's folder and tombstones the folder the target used to occupy — a durable standing order that the deletion is allowed to fail against, retried at the next boot.

Boot-time folder classification replaces "every directory under the storage root is a catalog": a folder is adopted only on positive evidence, and one evitaDB abandoned is now distinguishable from one an operator placed deliberately.

## Crash safety

The commit window — WAL appended, matching bootstrap not yet written — is recovered by forward replay, now implemented for the rename/replace and state-transition operators. Three defects found and fixed along the way, each with a test that fails without it:

- **A boot that recovered a crash bricked the next one.** Truncation took its cut position from the replayed snapshot, whose log offsets are the pre-crash ones, and so cut away the very record the replay had just recovered. Engine-wide rather than specific to this work — it affected every operator implementing replay — and only reachable when nothing committed between the two boots, which is why it had never been hit.
- **A catalog nobody had opened a session on could not be quiesced**, so requests were served against a catalog while it was being replaced (measured: 850 sessions).
- **A renamed catalog's old name answered "busy" forever** instead of "not found", because the session registry was never moved to the surviving name.

A crash also used to leave the `.catalogname` label — the only artefact that makes a bare storage directory readable — naming the folder's previous occupant permanently. It now converges when the catalog is loaded. The folder a replace supersedes is reclaimed by the boot that recovers, rather than the one after it.

## Testing

Full functional suite: **21176 tests, 0 failures**. The two errors are environmental — `ExportS3ServiceTest` needs Docker, and one gRPC client test lost its connection during dataset build under load (64/64 green re-run in isolation).

Every fix ships with a test that was verified to fail without it, one counterfactual at a time.

## Notes for review

- Decision record: `documentation/adr/2026-08-06-catalog-folder-decoupling.md`
- Protocol and invariants for engine-state operations: `documentation/developer/engine-state-operations.md`
- Known gap, recorded rather than fixed: the boot inventory divergence is computed before forward replay and drained after it, so a catalog the replay just re-registered can be marked missing on a stale reading. The engine's own operations cannot produce it; an external folder deletion during an unrecovered commit window can.


novoj:
## CodeQL: the Zip Slip alert on `RestoreTask.java:158` is a false positive

**Alert:** *Arbitrary file access during archive extraction ("Zip Slip") — unsanitized archive entry, which may contain `..`, is used in a file system operation.* High severity, reported as a **new** alert in code changed by this PR.

### The guard is present and it works

```java
final String fileName = getFileNameWithCatalogRename(entry.getName(), catalogName);
final Path entryPath = storagePath.resolve(fileName).normalize();
Assert.isTrue(entryPath.startsWith(storagePath), "Bad ZIP entry!");
try (final FileChannel fileChannel = FileChannel.open(entryPath, CREATE_NEW, WRITE)) {
```

This is the textbook containment check — resolve, `normalize()` to collapse any `..`, then refuse anything that escaped the destination root. `Assert.isTrue` throws `EvitaInvalidUsageException` when the condition is false (`evita_common/.../utils/Assert.java:115`), so a malicious entry aborts the restore before `FileChannel.open` is reached.

CodeQL does not model a project-local assertion helper as a sanitizer, so it sees the guard as a no-op and reports the path as unprotected. Rewriting it as an inline `if (...) throw` would silence the query — that is the only thing such a rewrite would change.

Worth noting the check also fails *closed* in the one asymmetric case: `entryPath` is normalized while `storagePath` is not, so a storage root containing a `.` or `..` segment makes `startsWith` return false and the restore aborts, rather than letting anything through.

### Why it surfaced now

The guard is byte-identical to `dev`, on the same lines. CodeQL reports alerts in code *changed by the PR*, and this PR rewrote the surrounding lines — `storagePath` is now `storageDirectory.resolve(catalogFolderId.id())` from the folder allocator instead of `storageDirectory.resolve(catalogName)`. The extraction loop and its containment check were not touched, and the vulnerability surface is unchanged.

### Requested action

Dismiss as **false positive**. If it becomes noisy on future PRs that touch this file, the durable fix is to make the guard legible to the query — an explicit `if (!entryPath.startsWith(storagePath)) throw ...` — rather than re-dismissing each time.

PR #1439, opened 2026-08-24
fix: keep chain index leaf page stamping create-free during the commit flush

### LINKED ISSUE #1437: Catalog upgraded to 2026.2 is marked CORRUPTED on first write to a legacy ChainIndex
## Summary

A catalog upgraded from an older storage protocol to `2026.2` (protocol 6) is marked **CORRUPTED** on the first write that touches a `ChainIndex` restored from a legacy, non-paged `ChainIndexStoragePart`:

```
ERROR TransactionManager - Error while committing transaction: 22072.
io.evitadb.exception.GenericEvitaInternalError: Transaction is already committed / rolled back,
  no new transactional memory layer may be created at this time!
	at io.evitadb.core.transaction.memory.TransactionalLayerMaintainer.getOrCreateTransactionalMemoryLayer(TransactionalLayerMaintainer.java:232)
	at io.evitadb.index.array.UnorderedLookupTree$LeafNode.setPageSequence(UnorderedLookupTree.java:2767)
	at io.evitadb.index.page.PageStreamRegistry.collectChangedPages(PageStreamRegistry.java:343)
	at io.evitadb.index.attribute.ChainIndex.doAppendStorageParts(ChainIndex.java:962)
	at io.evitadb.index.attribute.AttributeIndex.getModifiedStorageParts(AttributeIndex.java:1314)
	at io.evitadb.core.buffer.DataStoreChanges.popTrappedUpdates(DataStoreChanges.java:296)
	at io.evitadb.core.collection.EntityCollection.flush(EntityCollection.java:1936)
	at io.evitadb.core.catalog.Catalog.flush(Catalog.java:1918)
	at io.evitadb.core.transaction.TransactionTrunkFinalizer.commitCatalogChanges(TransactionTrunkFinalizer.java:108)
```

This was observed on a production dataset (`decodoma_pl`) after installing `2026.2`. The storage-protocol migration itself (`Migration_2026_2`) completed correctly — the catalog header sits at protocol 6 and the catalog **loads** without error. The failure happens afterwards, in the trunk-incorporation flush of the first transaction (in the observed case, the replay of a single pending WAL transaction at boot).

## Root cause

`PageStreamRegistry.collectChangedPages` assigns a page sequence to every not-yet-paged leaf and stamps it onto the live node:

```java
if (freshLeaf) {
    pageSequence = allocate(streamId);
    handle.setPageSequence(pageSequence);   // <-- create-on-write
    anyFreshLeaf = true;
}
...
if (freshLeaf || handle.isDirty()) {
    changedPages.add(pageBuilder.build(pageSequence, handle));
    handle.clearDirty();                    // <-- create-on-write
}
```

`UnorderedLookupTree$LeafNode.setPageSequence` / `clearDirty` route through
`Transaction.getOrCreateTransactionalMemoryLayer(this)`. This collection runs **inside**
`TransactionTrunkFinalizer.commitCatalogChanges` → `Catalog.flush(...)`, which happens **after**
`TransactionalLayerMaintainer.commit()` has already set `allowTransactionalLayerCreation = false`.
For a leaf the transaction never touched there is no existing diff layer, so the call tries to create
one and trips the premise.

A **legacy chain index** is exactly that situation. `AttributeIndexLoader.fetchChain` restores a
non-paged `ChainIndexStoragePart` via `new ChainIndex(referenceKey, attributeIndexKey, chains, elementStates)`,
which builds the element tree by ordinary insertion — so **every** leaf carries
`UNASSIGNED_PAGE_SEQUENCE` and every leaf is "fresh" at the first transactional flush, including the
ones the transaction never touched. Instrumented run on the production dataset:

```
chain flush entityIndexPk=1 key=AttributeIndexKey[referenceName=null, attributeName=order, locale=null]
            leaves=3 freshLeaves=3 livePages=[]
```

The codebase already documents this exact hazard for the sibling path: `LeafNode.forgetPage()`
exists solely because `forgetPageStream()` "runs at flush time … after `commit` has forbidden new
layer creation" and "visits EVERY leaf including collapse survivors this transaction never touched".
The stamping path was never given the same create-free treatment. Every other paged tree
(`TransactionalElementBPlusTree`, `TransactionalBucketBPlusTree`, `TransactionalLongBPlusTree`)
writes `pageSequence` directly to the field; `UnorderedLookupTree` is the only one that does not.

## Why it does not reproduce on a natively-created 2026.2 catalog

- warm-up (bulk) flushes run without a transaction, so the stamp writes straight to the node;
- after go-live every leaf already carries a page sequence;
- a split-born leaf acquires its diff layer during the split's own mutation
  (`getRecordIdsForUpdate` / `setCount`), while layer creation is still permitted.

It needs a legacy (pre-paging) chain index whose element tree spans more than one leaf.

## Impact

- Any catalog created before `2026.2` that has a chained (`Predecessor` / `ReferencedEntityPredecessor`)
  attribute whose chain-element tree spans more than one leaf.
- Fires on the **first write** touching that entity index — WAL replay at boot *or* an ordinary commit;
  both go through `TransactionTrunkFinalizer`.
- Only the first such entity index fails; the remaining ones stay latent until they are written.
- Present in `v2026.2.0` … `v2026.2.4` and on `release_2026-2` HEAD.

Note that a migration-side fix cannot repair an already-upgraded catalog: its header is at protocol 6,
so `Migration_2026_2` never runs again. The fix has to live in the index/tree layer.

## Fix

Make `UnorderedLookupTree$LeafNode.setPageSequence` and `clearDirty` create-free — write through an
existing transactional layer when one is present, otherwise write the committed baseline field in
place — mirroring `LeafNode.forgetPage()` and the B+ tree family. This is correct in both directions:
with a layer present the merge takes the layer value, and for an untouched leaf
`createCopyWithMergedTransactionalMemory` returns `this`, so the in-place stamp is what the merge
carries forward anyway.

Verified against the production dataset: the catalog loads, replays WAL transaction 22072 and reaches
`ALIVE`, and the chain index self-heals to the PAGED shape (`livePages=[0, 1, 2]`) on that same flush.

novoj:
Closes #1437

## Problem

A catalog upgraded to storage protocol 6 by `Migration_2026_2` loads fine, but goes `CORRUPTED` on its **first write**:

```
Transaction is already committed / rolled back, no new transactional memory layer
may be created at this time!
```

The migration itself is not at fault — bootstrap history on the reported dataset confirms it completed cleanly (protocol 4 → 6), and the catalog served reads afterwards. The failure is in the trunk-incorporation flush.

## Root cause

`TransactionTrunkFinalizer.commitCatalogChanges` flushes the catalog **after** `TransactionalLayerMaintainer#commit` has sealed layer creation. During that flush `PageStreamRegistry.collectChangedPages` walks *every* leaf of a paged tree and stamps each one not yet paged.

`UnorderedLookupTree$LeafNode.setPageSequence` and `clearDirty` routed that write through `Transaction.getOrCreateTransactionalMemoryLayer`. A leaf the transaction never touched carries no layer, so the create-on-write path tripped the "already committed" premise and aborted the commit.

Why it only bites upgraded catalogs: `AttributeIndexLoader.fetchChain` restores a legacy, non-paged `ChainIndexStoragePart` into a tree where **every** leaf is fresh, so the very first flush stamps leaves the write never touched. Native 2026.2 catalogs never hit it — warm-up flushes run without a transaction, and after go-live every leaf already holds a page sequence (split-born leaves acquire a layer during the split's own mutation).

The fix has to live in the tree, not the migration: an already-upgraded catalog is at protocol 6, so `Migration_2026_2` never runs again.

## Fix

Both methods now write through an **existing** layer only, never creating one — the same create-free contract `forgetPage()` already obeyed and the read side `isDirty()` already used. With no layer, `createCopyWithMergedTransactionalMemory` returns `this`, so writing the committed baseline in place is exactly what the merge would have carried forward. The sibling paged tree families (`TransactionalLongBPlusTree`, `TransactionalElementBPlusTree`, `TransactionalBucketBPlusTree`) were already create-free here — `UnorderedLookupTree` was the outlier.

## Verification

- **New regression test** `ChainIndexLegacyLoadTransactionalFlushTest` — builds a 3-leaf chain index the legacy way (no page stream), then models the trunk path: flush inside `commit()`, followed by `verifyLayerWasFullySwept()`. Reverting the fix reproduces the production error verbatim.
- **Real reported dataset** — the catalog reaches `ALIVE` and self-heals to `livePages=[0,1,2]`; a second boot loads clean in 8s with no WAL replay.
- **Full `evita_functional_tests`** — 20 914 tests. The only two reds are pre-existing and unrelated: `ExportS3ServiceTest` needs Docker (absent on the test machine), and `CdcCallbackDispatcherTest.shouldNotCreateThreadsWhenCallbacksAreRefused` asserts on the JVM-wide `getTotalStartedThreadCount()` and is inflated by concurrently running tests — green 4/4 in isolation.

### DIFF
--- a/evita_engine/src/main/java/io/evitadb/index/array/UnorderedLookupTree.java
+++ b/evita_engine/src/main/java/io/evitadb/index/array/UnorderedLookupTree.java
@@ -1044,7 +1044,7 @@ public class UnorderedLookupTree implements
 		for (final LeafNode leaf : collectLeaves()) {
 			// create-free reset: this runs at flush time after commit() has forbidden new layer creation, and the walk
 			// visits EVERY leaf including collapse survivors this transaction never touched (which carry no layer) - so
-			// it must NOT route through the create-on-write setPageSequence/clearDirty (see LeafNode.forgetPage)
+			// it must not create a layer either (see LeafNode.forgetPage)
 			leaf.forgetPage();
 		}
 	}
@@ -2759,12 +2759,26 @@ public class UnorderedLookupTree implements
 		}
 
 		/**
-		 * Stamps this leaf's logical persistence page sequence, decoupling into the transactional layer when present.
-		 * Structural bookkeeping only — does NOT flag the leaf dirty.
+		 * Stamps this leaf's logical persistence page sequence, writing through an EXISTING transactional layer when
+		 * one is present but NEVER creating one — the same create-free contract as {@link #forgetPage()}, and for the
+		 * same reason. Structural bookkeeping only — does NOT flag the leaf dirty.
+		 *
+		 * This is called from the flush path ({@code PageStreamRegistry.collectChangedPages}), which runs INSIDE the
+		 * commit — {@code TransactionTrunkFinalizer.commitCatalogChanges} flushes the catalog after
+		 * {@link io.evitadb.core.transaction.memory.TransactionalLayerMaintainer#commit} has already forbidden new
+		 * layer creation. The emitter walks EVERY leaf and stamps each one not yet paged, so a leaf the transaction
+		 * never touched (which carries no layer) reaches this method routinely: a create-on-write stamp would trip the
+		 * "already committed" premise on it and abort the commit. That is exactly what happens on the first write to a
+		 * chain index restored from a legacy, non-paged `ChainIndexStoragePart` — every leaf of such a tree is fresh
+		 * (see issue #1437).
+		 *
+		 * Writing the committed baseline field in place is what the merge carries forward anyway: with no layer,
+		 * {@link #createCopyWithMergedTransactionalMemory} returns `this`. When a layer DOES exist the merge takes the
+		 * layer's value, so the stamp must land there — hence the branch rather than an unconditional field write.
 		 */
 		void setPageSequence(int pageSequence) {
 			final LeafNode layer = this.transactionalLayer ?
-				Transaction.getOrCreateTransactionalMemoryLayer(this) : null;
+				Transaction.getTransactionalMemoryLayerIfExists(this) : null;
 			if (layer == null) {
 				this.pageSequence = pageSequence;
 			} else {
@@ -2782,12 +2796,15 @@ public class UnorderedLookupTree implements
 		}
 
 		/**
-		 * Clears this leaf's dirty flag, decoupling into the transactional layer when present. Called by the page emitter
-		 * once the leaf's page has been collected for the current flush.
+		 * Clears this leaf's dirty flag, writing through an EXISTING transactional layer when one is present but NEVER
+		 * creating one. Called by the page emitter once the leaf's page has been collected for the current flush —
+		 * i.e. from the very same post-{@code commit()} flush pass as {@link #setPageSequence(int)}, and therefore
+		 * bound by the same create-free contract (see that method for the full rationale). Symmetric with the
+		 * create-free read path {@link #isDirty()}: the flag is cleared exactly where it was read from.
 		 */
 		void clearDirty() {
 			final LeafNode layer = this.transactionalLayer ?
-				Transaction.getOrCreateTransactionalMemoryLayer(this) : null;
+				Transaction.getTransactionalMemoryLayerIfExists(this) : null;
 			if (layer == null) {
 				this.dirty = false;
 			} else {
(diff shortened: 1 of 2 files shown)


novoj:
Superseded by #1452, which merges this branch unchanged into the combined `hotfix-2026-2-5` batch targeting `master`. No content was rebased or squashed — the commits from this PR's head branch appear verbatim in #1452.

Issue #1388, opened 2026-08-05
Long-lived streaming driver calls abort after exactly 15 s (goLive, transaction commit, backup progress)


novoj:
# Long-lived streaming driver calls abort after exactly 15 s (`goLive`, transaction commit, backup progress)

**Labels:** `bug` · **Affects:** v2026.2.2 and current `dev` / `release_2026-2` · **Introduced by:** `c9e72b8c4` (#1367)

## Symptom

`EvitaSessionContract.goLiveAndClose()` against a large catalog throws

```
CompletionException: io.grpc.StatusRuntimeException: DEADLINE_EXCEEDED: deadline exceeded after 3600000000000ns
  Caused by: com.linecorp.armeria.client.ResponseTimeoutException
    at com.linecorp.armeria.internal.common.DefaultCancellationScheduler.invokeTask(...:362)
```

…while the server logs `Catalog 'senesi' is now alive!` seconds later and the catalog is complete and
correct. The client reports a hard failure for an operation that **succeeded**.

The advertised deadline is a red herring: it names the *gRPC* deadline (3600 s here), but what fired
is Armeria's *response* timeout. An error claiming a one-hour deadline expired after fifteen seconds
costs whoever hits it an hour of misdirection.

## It is not a `goLive` bug — five streaming calls are affected

Every site in `EvitaClientSession` that re-arms the response timeout per streamed message is capped
at 15 s from call start under default configuration:

| line | call | exposure |
|---|---|---|
| 508 | `goLiveAndCloseWithProgress` | WARM_UP → ALIVE on a large catalog |
| **593** | **`closeNow(CommitBehavior)`** | **transaction commit** |
| **720** | **`closeNowWithProgress()`** | **transaction commit with progress** |
| 3026 | `MutationsStreamObserver` (`getMutationsHistory`) | CDC / mutation-history streaming |
| 3182 | `BackupProgressObserver` | backup / restore progress |

**A transaction commit that takes over 15 s to reach `WAIT_FOR_CHANGES_VISIBLE` fails identically** —
a far more common production path than a bulk reindex, and one where "reported failed, actually
committed" is the worst possible outcome.

## Root cause

`c9e72b8c4` changed `EvitaClient` so Armeria's `RetryingClient` decorator is installed
**unconditionally**; previously it was installed only when `configuration.retry()` was `true`, and
that flag defaults to `false`. So with default config the decorator was absent before and is present
now.

From the Armeria 1.40.0 sources:

[...]

PR #1450, opened 2026-08-24
fix: pace gRPC server-streaming producers and give streams their own timeout budget

### LINKED ISSUE #1441: gRPC server-streaming has no flow control: fetchFile buffers the whole file for a slow client
## Summary

gRPC server-streaming RPCs push messages into Armeria's outbound queue without ever consulting transport
readiness. The queue is unbounded, so a consumer slower than the producer causes the server to materialise
the entire payload in memory. `FetchFile` is the acute case: downloading a large backup over a slow link
(e.g. a `kubectl port-forward` tunnel) kills the RPC part-way through with a bare `UNKNOWN` status and no
message.

A second, independent defect makes the standard fix a no-op until it is repaired: `ObservabilityInterceptor`
decorates the `ServerCall` without forwarding `isReady()`.

## Status

**Steps 0–3 are implemented and verified.** Retained memory for a 512 MiB download to a stalled client
went from **1035 MiB to 14.4 MiB**; 648 functional tests (`driver | (grpc & external_api)`) and 4
long-running tests are green.

Step 4 (CDC demand gating) is **split out into #1446** (milestone 2026.3) — a subscriber whose WAL
pointer has been purged stalls silently and permanently, and gating would make that the normal
consequence of a slow consumer. That has to land first.

Two size caps found along the way — both pre-existing, both independent of the buffering defect, both
fixed — are listed under *Follow-ups*.

Design rationale, rejected options and the invariants a future streaming RPC must preserve are recorded in
`documentation/adr/2026-08-24-grpc-streaming-backpressure-readiness-gate.md`. The rest of this issue is the
original diagnosis, kept as the evidence trail.

One finding worth repeating outside the ADR, because it silently invalidates tests: **the default test
engine collapses the request pool into a direct executor** (`EvitaParameterResolver` builds `Evita` with
`directExecutor = !useRealThreadPools`), so `executeWithClientContext` runs the streaming loop *inline on
the Armeria event loop*. Readiness gating is impossible there by construction — the gate detects it, warns
and runs ungated rather than self-deadlocking. Any test that means to cover back-pressure, ordering, or
"the event loop stays free" must declare `@DataSet(..., useRealThreadPools = true)`, or it measures the
opposite topology and can never fail.

## Symptom

A 690 MB backup download through evitaLab terminates at roughly 250 MB. The gRPC-Web response body arrives
complete with a trailer frame carrying `grpc-status: 2` (`UNKNOWN`) and an empty `grpc-message` — surfacing
client-side as `ConnectError: [unknown]` with `Caused by: undefined`. The same download succeeds over a fast
link.

### 1. `fetchFile` streams without flow control

`EvitaManagementService.java:721-727` reads the file in a tight loop and calls `responseObserver.onNext(...)`
for every 64 KB with no `ServerCallStreamObserver.isReady()` / `setOnReadyHandler` gating.

Armeria's `StreamingServerCall.doSendMessage` marshals each message eagerly via `toPayload(...)` and
`tryWrite`s the result into a `DefaultStreamMessage`, whose backing store is a plain `Queue<Object>` — its
only capacity field is `INITIAL_CAPACITY`, a starting size, not a bound. `tryWrite` returns `false` only when
the stream is already closed, never for capacity. There is no backpressure to be had except through
`isReady()`, which in Armeria is `!isCloseCalled() && pendingMessages == 0`.

Note this is **not** how grpc-go or the gRPC C++ sync API behave, where `Send`/`Write` blocks on the
flow-control window. In grpc-java the send path is asynchronous and never blocks; `isReady()`/
`setOnReadyHandler` exist precisely because of that.

### 2. `ObservabilityServerCall` does not forward `isReady()`

`ObservabilityInterceptor.java:104` — `ObservabilityServerCall` extends `ServerCall` directly rather than
`ForwardingServerCall`, and overrides only `request`, `sendHeaders`, `sendMessage`, `close`, `isCancelled`
and `getMethodDescriptor`. `io.grpc.ServerCall.isReady()` is **not** abstract — in grpc-api 1.83.1 it
compiles to `iconst_1; ireturn`, an unconditional `true`.

It is the innermost decorator (Armeria's `ServerInterceptors.intercept` makes the first-registered
interceptor outermost), so it is what a service method's `ServerCallStreamObserver` actually queries.
Meanwhile `ObservabilityListener` *does* extend `ForwardingServerCallListener`, so `onReady()` forwards
correctly — which makes it worse, not better: the on-ready callback fires while `isReady()` answers `true`
forever, and a naive `while (isReady())` loop runs straight through the whole file.

`setOnReadyThreshold(int)` is likewise not forwarded (and Armeria ignores thresholds anyway). The same
omission also leaves `getAttributes()`, `getAuthority()`, `getSecurityLevel()`, `setCompression` and
`setMessageCompression` on their `io.grpc.ServerCall` defaults.

### 3. The buffered bytes are off-heap, and cost ~2x the payload

Because marshalling is eager, the queued objects are pooled Netty `ByteBuf`s, not the protobuf messages.
Measured on a **512 MiB** file with a stalled client:

| counter | before | while stalled | growth |
|---|---|---|---|
| heap | 70.3 MB | 76.9 MB | **6.2 MiB** |
| Netty direct (`PlatformDependent.usedDirectMemory`) | 4.4 MB | 1083.6 MB | **1029.2 MiB** |
| NIO direct (`BufferPoolMXBean`) | 38.0 MB | 1117.3 MB | **1029.3 MiB** |

The amplification is **growth against the payload**: 1029.2 MiB of direct memory to hold a 512 MiB file,
i.e. **2.01x**, or ~128.6 KiB of direct memory per 64 KiB chunk (8192 chunks). The two independent
counters agree to within 0.1 MiB. (The absolute readings above are *not* the ratio — direct memory rose
271-fold from its idle baseline, which is a different number and not the interesting one.)

The mechanism is pooled-allocator overhead rather than a straight copy: Netty's counter tracks memory
**reserved from the OS** (arena chunks) rather than live bytes, so size-class rounding of the
65,541-byte payload (64 KiB + the 5-byte gRPC frame header) and arena reservation/fragmentation both
contribute. The exact split between them has not been attributed and would need an allocator dump; the
2.01x figure itself is measured and reproducible to within 40 KB across runs.

The practical consequence is that the ceiling such a download hits is `-XX:MaxDirectMemorySize`, not
`-Xmx`, and that a 690 MB file needs roughly 1.4 GB of direct memory to buffer.

### Why the failure surfaces as bare `UNKNOWN`

`StreamingServerCall.sendMessage` posts `doSendMessage` to the Armeria event loop, so the allocation that
finally fails does so inside Armeria, whose fallback for an escaping `Throwable` is `UNKNOWN` with no
description (`verboseResponses` is not enabled in `GrpcProviderRegistrar`). Nothing in evitaDB remaps it
either: `EvitaService.executeWithClientContext` catches only `RuntimeException`, and
`ObservableThreadExecutor.ObservableRunnable.run` catches only `Exception` — an `Error` passes through both.

### Ruled out

Not the Armeria request timeout. `ApiOptions.DEFAULT_REQUEST_TIMEOUT` is 1000 ms and
`evita-configuration.yaml` sets `api.requestTimeoutInMillis: 2K`, but `FramedGrpcService.doPost` calls
`ctx.clearRequestTimeout()` for every framed request — which includes gRPC-Web. It does not apply to this
call.

## Evidence

Two tests were written to confirm/refute the diagnosis. Both are currently red, by design.

**`ObservabilityInterceptorReadinessTest`** (`evita_functional_tests`) — drives the interceptor with a
delegate call reporting `isReady() == false` and asserts the decorated call agrees:

```
expected: <false> but was: <true>
```

**`LongRunningGrpcFetchFileBackpressureTest`** (`evita_long_running_tests`) — a manual-flow-control gRPC-Web
client requests 4 chunks of a 512 MB file and then stops requesting; after a 20 s settle window the test
measures retained memory (heap plus off-heap) and asserts it stays below an eighth of the file:

```
Threads currently inside EvitaManagementService.fetchFile: 0
Client consumed 4 chunks (262144 B) of a 536870912 B file; server retained 1085766607 B (1035 MB)
  before:        heap=70337768  nettyDirect=4440351     nioDirect=38011251
  while stalled: heap=76857272  nettyDirect=1083621918  nioDirect=1117258354
```

Reproducible to within 40 KB across runs. The `0` on the first line is the decisive part: the server-side
streaming loop had already run to completion and closed the file while the client had taken 256 KB. After
the client resumes, all 512 MB arrive with a matching CRC32.

**After the fix**, the same test (now with 1 MB chunks and `useRealThreadPools = true`) reports:

```
Threads currently inside EvitaManagementService.fetchFile: 1
Client consumed 4 chunks (4194304 B) of a 536870912 B file; server retained 15070587 B (14 MB)
After the client drained the stream, retained 12614095 B relative to the start
```

**1035 MiB → 14.4 MiB**, of which 12.0 MiB is still present after the client drains — that is Netty arena
growth, not queued messages, so genuine in-flight data is ~2.4 MiB, about two chunks. The thread count
flipping from `0` to `1` is the other half of the proof: the producer is now parked in the gate rather
than finished. Two runs agreed to within 90 KB. The test's ceiling was retightened from 64 MB to 32 MB
accordingly — evidence on both sides now, rather than only on the failing one.

## Scope — other affected methods

`setOnReadyHandler` appears **zero** times in the entire gRPC server module. The 11 existing
`ServerCallStreamObserver` casts are all for `setOnCancelHandler`. Of the 20 server-streaming RPCs:

| RPC | Site | Risk |
|---|---|---|
| `FetchFile` | `EvitaManagementService.java:721-727` | **High** — whole file at disk speed |
| `GetMutationsHistory` / `GetMutationsHistoryForward` | `EvitaSessionService.java:2448-2456` | Medium — unbounded by history size |
| `GetTrafficRecordingHistory` | `EvitaTrafficRecordingService.java:170-176` | Medium |
| `RegisterChangeCatalogCapture` / `RegisterSystemChangeCapture` | `AbstractChangeCaptureSubscriber.java:260,276` | Lower — see below |
| 12 × `*WithProgress` | — | None — a handful of messages each |

CDC is lower risk than it looks: `DefaultChangeCaptureSubscription` already holds a bounded
`ArrayBlockingQueue` and emits one capture per `request(1)`, so it can only outrun a slow client at
*publisher* rate, not at disk speed.

### Separately: blocking work on the event loop

`GrpcProviderRegistrar` never calls `useBlockingTaskExecutor(true)`, and `FramedGrpcService` only uses
`ctx.blockingTaskExecutor()` when that flag (or the per-method `@Blocking` annotation) is set. So every gRPC
handler and listener callback starts on the event loop, and `executeWithClientContext` is what moves it off.

A scan of all four services found exactly one method that never reaches a hand-off:
**`EvitaManagementService.restoreCatalog(StreamObserver)`** (`:342-422`), the client-streaming upload
counterpart of `fetchFile`. On the event loop it performs `Files.createTempFile` / `Files.newOutputStream`
(`:346-352`), `backupFile.writeTo(outputStream)` per inbound chunk to an **unbuffered** stream (`:364`),
and `close` / `Files.size` / `Files.newInputStream` in `onCompleted` (`:392-398`). A 690 MB restore is
thousands of blocking write syscalls on an event loop thread.

`restoreCatalogUnary` is already correct (hands off to `transactionExecutor`); the remaining ~60 unary
methods across the four services all hand off correctly.

## Proposed changes

Ordered by dependency — nothing in step 1 works before step 0.

### 0. Make `ObservabilityServerCall` forward (prerequisite)

Change it to extend `ForwardingServerCall.SimpleForwardingServerCall<M, R>`, keeping the overrides that do
real work. Also restores the five other methods currently stuck on `io.grpc.ServerCall` defaults.

### 1. Gate the streaming producers on transport readiness

A small shared helper in the gRPC server module, used by `fetchFile`, `streamMutationsHistory` and
`getTrafficRecordingHistory`:

```java
final GrpcOutboundGate gate = GrpcOutboundGate.attach(serverCallObserver);
...
while ((bytesRead = inputStream.read(buffer)) != -1) {
    gate.awaitReady();              // parks this worker until isReady(), cancel, or timeout
    responseObserver.onNext(response);
}
```

This preserves the straight-line loop shape rather than forcing a callback state machine, and pins no thread
that is not already pinned — all three RPCs already hold a request-executor thread for their whole duration.

**Constraint that dictates the shape:** gRPC's `ServerCallStreamObserverImpl` rejects `setOnReadyHandler` /
`setOnCancelHandler` once the service method has returned, which is exactly what happens on the worker
thread `executeWithClientContext` hands off to — the codebase already documents this for cancel handlers at
`EvitaSessionService.java:2551-2554`. So `attach(...)` must be called synchronously in the service method,
**before** `executeWithClientContext`; only the loop goes to the worker.

`awaitReady` must wake on `setOnCancelHandler` and carry a bounded timeout, so a vanished client cannot pin
a worker thread indefinitely.

### 2. Raise `fetchFile`'s chunk size in the same change

Armeria's `isReady()` is `pendingMessages == 0`, so a gated loop allows exactly one message in flight. At the
original 64 KB that is ~8192 sequential event-loop round trips for a 512 MB file, and it can make
*fast*-client downloads slower than today. The binding ceiling is the receiver's max inbound message size —
4 MB in gRPC-Java, unlimited in Armeria's gRPC client, and evitaDB sets no explicit limit on either side.

**Done: 1 MB** (`EvitaManagementService.FETCH_FILE_CHUNK_SIZE`). This was the one change with a plausible
regression, so it was verified against the existing suites rather than the new tests: 372
`grpc & external_api` and 276 `driver` functional tests are green, including the end-to-end
backup → `fetchFile` → streaming `restoreCatalog` round trip.

### 3. Move `restoreCatalog`'s blocking work off the event loop

Take the temp-file creation, the per-chunk write and the finalisation off the event loop, and wrap the
output stream in a `BufferedOutputStream` (it is currently unbuffered).

Today's inbound throttling is *accidental*: gRPC auto-requests the next message only after `onMessage`
returns, so the inline blocking write is what limits the client. Moving the write off-thread must therefore
replace it explicitly — `disableAutoRequest()`, then `request(1)` when each write completes, with the writes
on a **single serialized task chain per call**.

**Rejected: annotating the method `@Blocking`.** Armeria supports it per-method
(`HandlerRegistry.Builder.needToUseBlockingTaskExecutor` looks for
`com.linecorp.armeria.server.annotation.Blocking`) and it would be a one-line fix routing the handler and its
listener callbacks to `evita.getServiceExecutor()` (already wired at `ExternalApiServer.java:571`). Not
recommended: `AbstractServerCall` dispatches each message as `blockingExecutor.execute(() -> invokeOnMessage(...))`
onto a **shared pool**, so per-call ordering would rest on an undocumented implicit invariant — that
auto-request never lets a second message be dispatched before the first returns. If that ever fails the
symptom is a silently corrupted catalog ZIP. Worth revisiting only if Armeria documents the guarantee.

### 4. CDC — gate engine mutations only, and not yet

Both shared publishers have two consumption paths: an in-memory ring buffer for a subscriber keeping up, and
a WAL fallback via `RingBuffer.OutsideScopeException` once it falls behind
(`ChangeCatalogCaptureSharedPublisher.java:62-63`, `ChangeSystemCaptureSharedPublisher.java:76-77`). Both are
pull-based, so withholding demand loses nothing — it just moves the subscriber from the ring to the WAL.

There is a third path that is neither: `processHostEvent` dispatches `HostSystemEvent`s through
`DefaultChangeCaptureSubscription.deliverImmediate` (`:306-319`), which **silently drops** the event when
`requested == 0` — documented as live-tail-only with no replay. Gating demand there would widen that drop
window from microseconds to however long the client lags, turning buffering into silent event loss. Host
events must stay ungated.

**This step is blocked — see below.**

## Blocking finding: falling outside WAL retention fails silently

Gating CDC demand makes subscribers fall out of the ring buffer and onto the WAL path far more often. That
path has a pre-existing hole:

- `AbstractMutationLog.findWalIndexFor(long)` returns **-1** when the requested version is in no retained WAL
  file — i.e. after `removeWalFiles()` has purged it (`walFileCountKept`, default **8**).
- That -1 goes straight into `createSupplier` (`:1994`) and on to `AbstractMutationSupplier`, whose
  constructor resolves a filename that does not exist, hits `!this.walFile.exists()` (`:226`) and leaves
  `observableInput`/`transactionMutation` null.
- `MutationSupplier.get()` then returns `null` (`:140`) and the stream is **empty**.

So a subscriber whose pointer has been purged receives an empty stream with no exception, no error to the
client and only a DEBUG-level log line. Its `lastVersion`/`lastIndex` never advance, so it re-requests the
same purged pointer forever: a CDC stream that looks alive (heartbeats keep flowing) but silently delivers
nothing, permanently.

This is a pre-existing defect, not one introduced here — but step 4 makes it far easier to reach, so it
should be fixed (or at minimum made loud) first. It probably warrants its own issue.

## Backward compatibility

No proto changes in steps 0–3, so no client regeneration and nothing required of evitaLab. Nothing touches
`evita_api` or the driver's public surface — no deprecations, and no Kryo/`serialVersionUID` implications.

| Step | Wire | Java API | Behaviour |
|---|---|---|---|
| 0 | unchanged | internal | Safe. No call sites for `isReady`, `setCompression` or `setMessageCompression` exist in the module, so nothing starts compressing or behaving differently |
| 1 | unchanged | internal | Same messages, order and count. Risk is throughput, see step 2 |
| 2 | unchanged | none | Safe below gRPC's 4 MB default. Older clients simply receive fewer, larger messages; chunking is not part of the contract, `totalSizeInBytes` is the whole file on every message |
| 3 | unchanged | internal | IO errors surface asynchronously rather than synchronously, so a client may push further chunks before learning. Response contract unchanged |
| 4 | unchanged | internal | Lossless for engine mutations; host events must be exempted; shifts stalled subscribers from ring to WAL reads |

The throughput risk in step 1 is the one to watch: `EvitaClientManagement.fetchFile` awaits the whole
download under the driver's configured timeout and throws `EvitaClientTimedOutException` on expiry, so a
slower download could newly time out for clients that pass today.

## Follow-ups

- ~~**Threshold calibration.**~~ Done — 32 MB, set from measurements on both sides (14.4 MB green,
  1035 MB red), with the 12.0 MB post-drain allocator floor accounted for.
- ~~**A long-running upload counterpart** for step 3.~~ Done —
  `LongRunningGrpcRestoreCatalogUploadTest` pushes a multi-megabyte backup through streaming
  `restoreCatalog` across hundreds of messages and verifies every restored entity's payload. It exists
  because the functional round trip backs up a ten-product catalog whose archive fits in a **single**
  64 KB message, so it never exercised the per-chunk hand-off at all. The test asserts its own chunk
  count so it cannot silently decay back into a duplicate of the functional one.
- **Found while writing the upload test — the driver could not download a file over 10 MiB.** Armeria
  defaults `maxResponseLength` to 10 MiB and counts the *whole* HTTP body, which on a server-streaming
  call is every message added together. `EvitaClient` never overrode it, so `fetchFile` died part-way
  through with `RESOURCE_EXHAUSTED`. Fixed: the cap is lifted on the streaming and CDC channels and
  kept on the unary one. Independent of the buffering defect, same user-visible symptom.
- **Also found and fixed — the driver could not restore a backup over `api.maxEntitySizeInBytes`
  (2 MB default).** A client-streaming upload is a *single* HTTP request, so Armeria's
  `maxRequestLength` (`ExternalApiServer.java:617`) bounded the **whole backup** rather than a chunk;
  anything larger died part-way through with a bare `RESOURCE_EXHAUSTED`, making `EvitaClient`'s restore
  unusable for a catalog of any real size. The limit itself is deliberate operator policy and was left
  alone — raising it would override a guard that applies to REST and GraphQL too. Instead
  `EvitaClientManagement.restoreCatalog(name, size, stream)` now uploads through `RestoreCatalogUnary`
  (on the wire since 2024-09), where each 512 KB chunk is its own request and only the chunk has to fit.
  The upload stub deliberately sits on the **non-retrying** streaming channel: appends are not
  idempotent and the unary channel replays on timeouts and 503/504/UNKNOWN when `retry` is enabled.
  Clients still using the streaming RPC — gRPC-Web, third parties — now get `RESOURCE_EXHAUSTED` **with
  a description** naming the limit and pointing at the unary RPC, instead of a bare status.
  Verified: a 5.47 MB backup (2.6x the old ceiling) restored across ~10 chunks with all 200 entities'
  payloads compared byte for byte.
- **Closed: the client-streaming `RestoreCatalog` is covered again.**
  `LongRunningGrpcRestoreCatalogUploadTest#shouldRestoreCatalogUploadedThroughClientStreamingRpc` backs
  up a second, deliberately small catalog (549 KB — a client-streaming upload is one request, so it must
  stay *under* the limit), pushes it through the raw `EvitaManagementServiceStub` in ~17 messages,
  restores, and compares all 20 payloads byte for byte. Awaiting the task without sleep-polling turned
  out to be a non-problem: `EvitaClientManagement.createTask(TaskStatus)` is public, so the test converts
  the response's `GrpcTaskStatus` and gets a real tracker-backed future.

  **Calibration changed what we believe about this code.** The intended counterfactual — unchaining
  `submitRestoration` from `onCompleted` — does **not** fail the test. Demand for chunk N+1 is issued
  from inside the *completed* write step for chunk N, so the demand protocol alone both orders the writes
  and withholds half-close until the last one lands. The ordered chain and `disableAutoRequest()` +
  `request(1)` are **redundant**; removing either alone is unobservable, and removing both with a 2 ms
  widener still passed at 9 chunks (a failure to detect, not evidence of safety). What the test
  *demonstrably* guards is the demand protocol: delete the per-write `request(1)` and the upload stalls
  dead, failing the 30 s latch deterministically. The ordering invariant remains unverified, and the
  `@Blocking` rejection is correspondingly weaker than originally written — recorded in the ADR and in
  `RestoreCatalogUploadObserver`'s class JavaDoc.

  Deprecating the client-streaming variant stays open and deliberate: gRPC-Web uses the unary RPC and so
  does the driver, so it has no known first-party client.
- **Resume support (separate enhancement).** `GrpcFetchFileRequest` carries only `fileId`
  (`GrpcEvitaManagementAPI.proto:321-324`), so a client cannot resume a broken transfer — any retry restarts
  at byte 0. An additive `optional int64 offset = 2` is wire-compatible both directions, but *silently* so:
  an old server ignores it and streams from byte 0 with no error. Pair it with an echo field on
  `GrpcFetchFileResponse` so the client can confirm the offset was honoured.

novoj:
Closes #1441

Server-streaming RPCs pushed messages into Armeria's unbounded outbound queue as fast as they could produce them. `tryWrite` refuses a message only once the stream is closed, never because it is full, so a client slower than the producer made the server materialise the whole payload — as pooled *direct* buffers, roughly 2x the payload once arena reservation is counted. `setOnReadyHandler` appeared **zero** times across all 20 server-streaming RPCs.

Producers now pace themselves against `ServerCallStreamObserver.isReady()` through a shared `GrpcOutboundGate`, which parks the producing worker until the transport drains.

### What gating exposed

Tying a handler's lifetime to the client's consumption rate made two timeout defects real:

**The per-message re-arm was compounding.** `ServiceRequestContext.requestTimeoutMillis()` looks like "the configured timeout" and is neither that nor the time remaining — Armeria stores the timeout relative to request *start*, and `SET_FROM_NOW` writes `elapsed + newTimeout` into that same field (`DefaultCancellationScheduler#setTimeoutNanosFromNow0`, 1.40.0). Six sites fed that getter back into the next re-arm, so a 2 s budget became 2.1 s, then 2.3 s, then 2.6 s, growing without bound. The rolling stall window the design rests on had become an ever-receding horizon a stalled client never reaches — invisible, because it only ever made calls *survive*.

**`requestTimeoutInMillis` is the wrong shape for a stream.** It bounds a whole request, which suits a unary call. A download's duration is a function of file size and link speed, so a whole-request budget is a cap on what can be transferred at all. New `api.endpoints.gRPC.streamingRequestTimeoutInMillis` (default `300K`) bounds *silence* instead and doubles as the gate's stall timeout, so both mechanisms that can end a stalled stream agree on when one is dead.

**The driver had no timeout tiers.** `EvitaClientManagement` took `ClientTimeoutOptions.timeout` — the unary 5 s budget — for everything including `fetchFile`, so a default-configured `EvitaClient` could not download anything taking longer than 5 s. Every test missed it because harness clients set `.timeout(10, MINUTES)`. `EvitaClientChannel.TimeoutTier` now travels with the channel and is paired with each stub at construction, the same seam that already makes "streaming stub on the retrying channel" not compile (#1388).

### Also in this line of work

- `restoreCatalog` moved its blocking file IO off the event loop onto an ordered chain with explicit inbound demand; the driver uploads through `RestoreCatalogUnary` so `api.maxEntitySizeInBytes` no longer caps a whole backup.
- `ObservabilityServerCall` now extends `SimpleForwardingServerCall` — it previously answered `isReady()` `true` forever, which would have silently no-op'd any readiness loop written against it.
- `ExceptionUtils.findInCauseChain` replaces a hand-rolled walker that guarded only self-referencing causes and would spin forever on `A -> B -> A`.

### Verification

| | result |
|---|---|
| 512 MiB to a stalled client | **1035 MiB → ~15 MB** retained (heap + both direct counters) |
| 64 MiB download, 2000 ms server deadline | completes in **6.7 s**; **dies at 2.1 s** with the re-arm removed |
| Re-arm compounding | pinned by `GrpcTimeoutUtilTest` — both strategies run side by side on real contexts |
| Gate stall-race grace | calibrated by zeroing it: horizon read 30124 ms vs a 30000 ms stall timeout |
| Suites | 80/80 targeted functional, 5/5 long-running streaming, full reactor build green |

Reasoning, rejected options and the two paths deliberately left on the whole-request budget are recorded in `documentation/adr/2026-08-24-grpc-streaming-backpressure-readiness-gate.md`.

### Known and unconfirmed

A production report of *sibling* RPCs failing during a download is discussed in the ADR's Consequences. The leading explanation is bandwidth saturation of the shared connection rather than HTTP/2 window exhaustion, but it is recorded as **unconfirmed** — the failed calls' status codes would distinguish the mechanisms.

### DIFF
--- a/documentation/adr/2026-08-05-streaming-calls-must-not-be-retry-decorated.md
+++ b/documentation/adr/2026-08-05-streaming-calls-must-not-be-retry-decorated.md
@@ -1,7 +1,7 @@
 ---
 title: Never decorate a streaming gRPC channel with RetryingClient
 date: 2026-08-05
-updated: 2026-08-05 09:45
+updated: 2026-08-24 14:20
 status: accepted
 kind: fix
 issues: [1388]
@@ -9,7 +9,7 @@ prs: [1389]
 areas: [evita_external_api/evita_external_api_grpc/client/src/main/java/io/evitadb/driver]
 supersedes: []
 superseded-by: []
-relates: [2026-08-03-driver-connection-resilience, 2026-08-04-client-pool-fail-fast-and-cdc-channel-isolation]
+relates: [2026-08-03-driver-connection-resilience, 2026-08-04-client-pool-fail-fast-and-cdc-channel-isolation, 2026-08-24-grpc-streaming-backpressure-readiness-gate]
 ---
 
 # Never decorate a streaming gRPC channel with RetryingClient
@@ -238,6 +238,9 @@ through the rewired stubs.
   *Consequences*.
 - `2026-08-04-client-pool-fail-fast-and-cdc-channel-isolation` — the CDC channel/event-loop isolation this
   builder split sits alongside; both landed in PR #1389, which two agents staged together.
+- `2026-08-24-grpc-streaming-backpressure-readiness-gate` — the server-side half of the same call sites:
+  this record says the driver must not retry a stream, that one says the server must pace it against
+  transport readiness rather than pushing at disk speed.
 
 ## Timeline
 
--- a/documentation/adr/2026-08-24-grpc-streaming-backpressure-readiness-gate.md
+++ b/documentation/adr/2026-08-24-grpc-streaming-backpressure-readiness-gate.md
@@ -1,0 +1,480 @@
+---
+title: Pace gRPC server-streaming producers with a readiness gate, and unblock large file transfers
+date: 2026-08-24
+updated: 2026-08-24 19:40
+status: accepted
+kind: fix
+issues: [1441]
+prs: []
+areas:
+  - evita_external_api/evita_external_api_grpc/server/src/main/java/io/evitadb/externalApi/grpc/utils
+  - evita_external_api/evita_external_api_grpc/server/src/main/java/io/evitadb/externalApi/grpc/services
+  - evita_external_api/evita_external_api_grpc/server/src/main/java/io/evitadb/externalApi/grpc/services/interceptors
+  - evita_external_api/evita_external_api_grpc/client/src/main/java/io/evitadb/driver
+supersedes: []
+superseded-by: []
+relates: [2026-08-05-streaming-calls-must-not-be-retry-decorated]
+---
+
+# Pace gRPC server-streaming producers with a readiness gate that parks the worker
+
+Server-streaming RPCs used to push messages into Armeria's outbound queue as fast as they could
+produce them, without ever asking whether the transport could take another one. The queue is
+unbounded, so a client slower than the producer made the server materialise the entire payload in
+memory. `fetchFile` was the acute case: a 690 MB backup download over a slow tunnel died part-way
+through with a bare `UNKNOWN`. Producers now pace themselves against `ServerCallStreamObserver.isReady()`
+through a shared helper, `GrpcOutboundGate`, which parks the producing worker until the transport
+drains. `restoreCatalog`, the upload counterpart, moved its blocking file IO off the event loop and
+replaced the throttle it lost in the process with explicit inbound demand. Two size caps found along
+the way — both pre-existing, both making a large transfer fail with a status that explained nothing —
+were fixed on the driver side in the same line of work.
+
+## Why
+
+gRPC-Java's send path never blocks — a fact that is easy to get wrong, because grpc-go and the
+gRPC C++ sync API *do* block on the flow-control window. `StreamObserver.onNext` marshals the message
+and hands it to Armeria's `StreamMessage`, whose `tryWrite` refuses a message only once the stream is
+closed, never because it is full. `isReady()` is the only back-pressure signal there is, and nothing
+in the module consulted it: `setOnReadyHandler` appeared **zero** times across all 20 server-streaming
+RPCs.
+
+The cost is paid off-heap and is worse than the payload. Because marshalling is eager, the queued
+objects are pooled Netty buffers: a 512 MiB file measured **1035 MiB** of retained direct memory,
+about 2x the payload once pooled-arena reservation is counted. The ceiling such a download hits is
+therefore `-XX:MaxDirectMemorySize`, not `-Xmx`, and the failure surfaces as `UNKNOWN` with no message
+because the allocation that finally fails does so inside Armeria, whose fallback for an escaping
+`Throwable` is exactly that.
+
+Two constraints made the fix non-obvious, and both are the reason this record exists rather than a
+commit message:
+
+- **`ServerCall.isReady()` is not abstract.** It ships a default returning an unconditional `true`.
+  `ObservabilityInterceptor` decorated the call without extending `ForwardingServerCall`, so it
+  answered `true` forever — any readiness-driven loop written against it would have been a silent
+  no-op. Nothing about the service method reveals this.
+- **The producing loop's thread is not guaranteed to be a worker.** `executeWithClientContext` hands
+  off to `Evita.getRequestExecutor()`, which for an engine built with `directExecutor = true` is an
+  `ImmediateExecutorService` that runs the task inline — on the transport event loop.
+
+### Previous state
+
+`fetchFile` read the file in 64 KB chunks in a tight `while` loop, calling `onNext` per chunk with no
+gating (`EvitaManagementService.java:721-727` before this change). `getMutationsHistory` and
+`getTrafficRecordingHistory` had the same shape via `Stream.forEach`. This was invisible for as long
+as every consumer kept up, which is the normal case on a LAN.
+
+`restoreCatalog` performed `Files.createTempFile`, a per-chunk `writeTo` against an **unbuffered**
+stream, and `close`/`Files.size`/`newInputStream` directly on the event loop. It was never throttled
+deliberately: gRPC auto-requests the next message only after `onMessage` returns, so the inline
+blocking write *was* the throttle — accidentally, and at the price of thousands of blocking write
+syscalls on a thread shared with every other connection.
+
+## Options considered
+
+### Option A — park the producing worker on a readiness gate (chosen)
+
+A small shared helper attached synchronously in the service method; the producing loop calls
+`awaitWritable()` before each `onNext` and stops when it returns `false`.
+
+- **Pros:** keeps the loop's straight-line shape, so try-with-resources, tracing scope and error
+  handling stay in one method. Pins no thread the RPC did not already own for its whole duration.
+  One helper covers every call site.
+- **Cons:** holds a request-executor thread for the duration of a slow download rather than for the
+  duration of a fast disk read. Needs a stall timeout so a client that stops reading without hanging
+  up cannot pin that thread forever.
+
+### Option B — drive the loop from the on-ready callback (declined)
+
+Turn each producer into a pump re-entered by `setOnReadyHandler`, holding the source and buffer as
+call state.
+
+- **Pros:** parks no thread at all; the natural shape for a purely async server.
+- **Rejected because:** the on-ready callback runs on the Armeria event loop, so the pump would have
+  to bounce every chunk back to a worker anyway — and then needs its own serialization guard to stop
+  two callbacks from pumping concurrently. That buys back the thread at the cost of turning three
+  straight-line loops into three hand-written state machines, each with its own resource-lifetime and
+  tracing-context problem. Worth revisiting only if request-executor exhaustion from slow downloads
+  is ever actually observed.
+
+### Option C — annotate the handlers `@Blocking` (declined, for `restoreCatalog`)
+
+Armeria supports the annotation per method and it would route the handler and its listener callbacks
+to `evita.getServiceExecutor()` in one line.
+
+- **Pros:** one-line change; no explicit flow control to maintain.
+- **Rejected because:** `AbstractServerCall` dispatches each message as
+  `blockingExecutor.execute(() -> invokeOnMessage(...))` onto a **shared pool**, so per-call ordering
+  would rest on an undocumented implicit invariant — that auto-request never lets a second message be
+  dispatched before the first returns. If that ever fails, the symptom is a silently corrupted catalog
+  ZIP, discovered at restore time. Revisit only if Armeria documents the ordering guarantee.
+
+## Decision
+
+**Chosen: Option A**, with Option C explicitly rejected for the upload path in favour of an explicit
+ordered chain.
+
+The deciding driver was that a slow download already owns its worker thread for the whole transfer;
+gating changes how *long* it is held, not whether. That is a bounded, measurable cost, and it is
+bounded further by a stall timeout. Option B's cost — three bespoke state machines whose failure modes
+are resource leaks and lost tracing context — is unbounded in review effort and lands on every future
+streaming RPC someone adds.
+
+Option B wins the day request-executor exhaustion from concurrent slow downloads is observed in
+production. That is the trigger for an ADR superseding this one.
+
+## Key technical details
+
+- **`GrpcOutboundGate.attach(...)` must be called synchronously in the service method, before it
+  returns.** gRPC's `ServerCallStreamObserverImpl` freezes handler registration the moment the service
+  method returns and throws for any later `setOnReadyHandler`/`setOnCancelHandler`. Only the loop
+  belongs on the worker. The codebase already documents the same constraint for cancel handlers in
+  `AbstractChangeCaptureSubscriber`.
+- **The gate owns the call's cancel handler.** gRPC keeps only the last one registered, so cleanup
+  that service methods used to register themselves (closing the underlying `Stream`) is passed into
+  `attach(...)` instead.
+- **`awaitWritable()` checks cancellation *before* readiness, in every branch.** Armeria increments
+  its pending-message counter in `sendMessage` and only unwinds it once the payload is genuinely
+  consumed, which never happens for a cancelled call — so readiness stays false forever. A gate that
+  consulted readiness first would pin a worker for the full stall timeout every time a client pressed
+  cancel. Registering a cancel handler also stops gRPC throwing `CANCELLED` from `onNext`, so this
+  check is the only thing left that stops a producing loop from reading its whole source into a dead
+  call.
+- **Producing on the event loop is detected and degrades to ungated**, with one warning per call.
+  Waiting there is a self-deadlock, not back-pressure: the parked thread is the only one that could
+  make readiness true again. This is reachable only for an engine built with `directExecutor = true`
+  (embedded test runs); `EvitaServer` always uses real pools. The consequence for tests is in
+  *Verification* below.
+- **`fetchFile`'s chunk size is 1 MB, and that is load-bearing, not cosmetic.** Armeria's readiness
+  is `pendingMessages == 0`, so a gated loop keeps exactly one message in flight and every chunk costs
+  an event-loop round trip; at the original 64 KB a large backup would have been over ten thousand
+  sequential round trips and slower than the unbounded loop for clients that keep up. The ceiling is
+  the receiver's maximum inbound message size — 4 MB in gRPC-Java, unlimited in Armeria's client.
+- **`ObservabilityServerCall` now extends `SimpleForwardingServerCall`.** Forwarding by default is the
+  invariant: `isReady`, `setOnReadyThreshold`, `getAttributes`, `getAuthority`, `getSecurityLevel`,
+  `setCompression` and `setMessageCompression` all have non-abstract defaults that silently disagree
+  with the transport.
+- **Repairing readiness broke a hidden dependency on it.**
+  `GlobalExceptionHandlerInterceptor.handleException` guarded its `close(...)` with
+  `if (serverCall.isReady())`. That only ever read as an open/closed test because readiness was
+  hard-wired to `true`; once real, it would have swallowed the error status for any RPC that had
+  already sent data, leaving the client on a stream that is never closed. It now guards on
+  `isCancelled()`.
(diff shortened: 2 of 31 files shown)


novoj:
Superseded by #1452, which merges this branch unchanged into the combined `hotfix-2026-2-5` batch targeting `master`. No content was rebased or squashed — the commits from this PR's head branch appear verbatim in #1452.

PR #1483, opened 2026-08-31
Trigram substring index: accelerate attributeContains and attributeEndsWith

### LINKED ISSUE #1454: Trigram substring index: fast attributeContains / attributeEndsWith
Sub-line of #258 (fulltext support): accelerate the literal `attributeContains` / `attributeEndsWith` path — today a full scan over all distinct values of the attribute (marked `TOBEDONE JNO naive and slow` in `FilterIndex`) — with an opt-in **trigram → distinct-value-id** index. Fulltext stays tokenized + ranked; this keeps the literal path literal, and fast.

Design and evidence live in the ADR: `documentation/adr/2026-08-24-fulltext-search-lucene-vs-inhouse/prototypes/p8-trigram-substring-index.md` (measurements and settled forks in its §35).

### What the spike proved (real corpora: cnc 972K articles, evita-demo-dataset)

- **Speed**: medium-selectivity patterns 1,000–5,000× faster than the scan (e.g. 307 ms → 131 µs); never slower in any of 35 measured cells (worst case, a very common 3-letter pattern: 3.5× faster); nonexistent patterns answer in tens of nanoseconds. Results byte-identical to the engine scan in all correctness cells.
- **Memory**: opt-in per attribute; realistic opt-in set ≈ 184 MB heap on 972K articles (+25% over that catalog's attribute-index heap), ~25 MB without the near-unique `title`. Zero cost when not enabled.
- **Updates**: churn on an existing value costs zero index writes; a new distinct value ~1.4–8.2 µs.
- **False positives are negligible** (worst 0.36 per true match) → positional postings are never needed.

### Settled design (rejected alternatives + reasons in ADR §35)

- `trigram(63-bit packed NFD code points) → RoaringBitmap<valueId>`, membership only; patterns < 3 code points fall back to the scan; `startsWith` untouched (has its own fast path).
- Hosted **once, globally** — never replicated per reduced index (the census showed 20K+ reduced indexes on fan-out shapes).
- Value dictionary = id columns attached to the existing front-coded value tree (~8 B/value marginal); packed value blocks rejected (1.4× faster at 8× memory). Directory must reference **stable leaf ids**, never positions.
- Hybrid postings: sorted `int[]` ≤ T, Roaring above; T = 128 pending a quiet-box latency check.
- Key lookup: open-addressing long-keyed table rebuilt on load; persisted form stays sorted `long[]`.

### Public API (decided)

Folded into the filterable call so the invalid state is unrepresentable, following the `unique(AttributeUniquenessType)` precedent:

```java
.filterable(FilterIndexCapability.SUBSTRING)   // String/String[] only; refused on non-empty collections
```

Query language unchanged — the planner uses the index transparently when present. `searchable()` (fulltext) deliberately does **not** fold the same way; the discriminator is recorded in `prototypes/schema-design.md` §5.4.

### Note on the valueId

The stable valueId introduced here (id column in the shared value tree + per-tree allocator) is engine infrastructure with this index as its **first consumer, not its owner** — the future reduced-index/sort-index value-dedup line (ADR §34) needs the same ids without the trigram flag. The id column is therefore gated by "any consumer needs ids on this tree", not by the SUBSTRING capability.

### Work plan

Sequential increments on the linked branch `1454-trigram-substring-index`; none of the steps is independently shippable (a schema flag without an index, an index without a planner path), so they merge back to `258-fulltext-support` as **one PR when the line is complete**:

1. Schema capability + refusal on non-empty collections (8-layer mirror, BWC)
2. valueId column + per-tree persisted allocator (storage parts, WAL, Kryo BWC)
3. `TrigramIndex` component on `GlobalEntityIndex` incl. transactional maintenance + bucket-death hook
4. Translator/planner integration with formula-cache key contract + fallback
5. Reduced-index plans (trigram → valueIds → global postings → AND reduced)
6. Functional tests, user docs, and the quiet-box confirmation of the T=128 constant

#545 (case-insensitivity) shares the normalization seam but is deliberately **not** part of this line — it will be addressed separately.

novoj:
**Draft — work remains on this branch. Opened for early visibility, not for merge.**

Adds a trigram substring index that accelerates `attributeContains` and `attributeEndsWith`. An
inverted index sorted by whole values says nothing about the middle of them, so a substring query
today tests every distinct value of the attribute. That is fine at a few thousand distinct values
and becomes the slowest part of the query at hundreds of thousands.

## What it adds

- **A new opt-in schema axis.** `AttributeFilterAccelerator.SUBSTRING_SEARCH`, declared through
  `acceleratedFor(...)` / `acceleratedForInScope(...)`, with `nonAcceleratedFor(...)` /
  `nonAcceleratedForInScope(...)` to withdraw it. It is never switched on implicitly — it costs
  memory and write throughput, so it is always the schema author's call.
- **The index itself.** `trigram (63-bit packed long) -> RoaringBitmap<valueId>`, hosted only on
  `GlobalEntityIndex`; reduced-index plans are served from that one instance. Postings are a hybrid
  of `int[]` for small cardinalities and Roaring above the threshold.
- **A value-id column on the shared value tree** of the inverted index, so postings address values
  by a dense primitive id rather than by the value itself.
- **A selectivity gate**, so the accelerator is used only where it is expected to win, and the scan
  is retained everywhere else.

## Result equivalence

The accelerator never changes what a query returns. Trigram intersection produces *candidates*;
every candidate is verified against the stored bytes before it is admitted.

## Backward compatibility

Value ids are new state on an existing structure. Catalogs written by earlier versions are read
through `_2026_2`-suffixed read-only Kryo serializers, marked
`@Deprecated(since = "2026.3", forRemoval = true)`, which throw on write. An existing catalog cannot
turn the accelerator on in place; it needs a re-import.

## Documentation

- `documentation/user/en/use/schema.md` — a "Filter accelerators" section covering what an
  accelerator is, how to declare one, how the index works, where it can be declared, and how to
  decide which attributes deserve one.
- `documentation/user/en/query/filtering/string.md` — performance notes on the three affected
  constraints, including why `attributeStartsWith` needs no accelerator.
- `documentation/adr/2026-08-24-fulltext-search-lucene-vs-inhouse/` — the decision record, its
  rejected options, and the measurements behind the gate.

## Still to do

- Remaining verification passes over the rebased tree.
- Final review of the persistence increment.

Ref: #1454
Ref: #258


evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
```diff
@@ -999,6 +1779,27 @@ public InvertedIndex createCopyWithMergedTransactionalMemory(
...
+				merged.valueIdAllocator =
+					transactionalLayer.getStateCopyWithCommittedChanges(this.valueIdAllocator);
+				// the emission baseline is owner-resident bookkeeping and must survive the commit, or every commit
+				// would look like the high-water had changed and rewrite the root for nothing
+				merged.emittedNextValueId = this.emittedNextValueId;
+				committedTree.installValueIdMinter(merged.valueIdAllocator::allocate);
+				// the directory belongs to the version it was built against: building it here, once, makes it
+				// immutable for the lifetime of this committed index, so a reader holding an older version keeps
+				// resolving against that version's own directory and MVCC needs no diff layer for it
+				committedTree.rebuildValueIdDirectoryAfterMerge();
```
novoj: **Silent false negatives after `goLive`.** The merge always takes `rebuildValueIdDirectoryAfterMerge()` and then clears `valueIdDirectoryStale` on the next line, never consulting the flag to choose the full rebuild. The incremental variant skips every leaf whose instance token (`final id`) is unchanged, and warm-up writes mutate leaves in place without changing that token. That is exactly the case the method's own javadoc in `TransactionalBucketBPlusTree` declares invalid: *"using this variant after such a mutation would silently skip the leaf that changed"*.

Scenario: declare `SUBSTRING_SEARCH` on an empty collection, bulk-import in warm-up, run one accelerated read-only query mid-import (it builds the directory and records the then-current leaf tokens), import more values into those leaves without splitting them, `goLive` (which only seals the maps), then commit any transaction touching the tree before the next accelerated read. The values written after the mid-import build never enter the directory, the flag is now false so the lazy repair is disabled, and `recordsOfMatchingValueId` answers `NO_LOCATION` for each of them: false negatives until a restart or a write to that leaf. The same happens without a mid-import query for trees under 256 values admitted through reduced-index fan-out.

Fix shape: when `this.valueIdDirectoryStale` is set at merge time, call the full `rebuildValueIdDirectory()` here instead. Better still, track *in-place since last build* separately from transactional writes so the full rebuild is paid only when a leaf really changed in place.

evita_engine/src/main/java/io/evitadb/index/usage/SchemaCapabilityUsageRegistry.java
```diff
@@ -619,6 +620,11 @@ private static boolean declaresCapability(
 			// AttributeSchemaAccessor lets a filter reach a unique-only attribute on exactly that basis. Testing only
 			// the `filterable()` flag here would therefore prune a live capability the moment a schema was re-adopted
 			case FILTERABLE -> attributeSchema.isFilterableInScope(scope) || attributeSchema.isUniqueInScope(scope);
+			// the bare declaration, deliberately not conjoined with FILTERABLE: the two are separate rows because
+			// they are separately droppable, and the schema already refuses a capability without filterability, so
+			// re-testing it here would only hide a corrupt schema rather than report one
+			case SUBSTRING_ACCELERATED -> attributeSchema.getAcceleratorsInScope(scope)
```
novoj, line 626: **`SUBSTRING_ACCELERATED` usage is registered but never recorded.** No write path records an update against this row and no query path records a request: `EntityIndexLocalMutationExecutor#reportAttributeTouched` (line 957) remembers only FILTERABLE / SORTABLE / UNIQUE, `AttributeSchemaAccessor` (lines 488-492) maps only those three traits, and `AbstractAttributeStringSearchTranslator` (line 371) selects the trigram index without any usage call.

Result: declare the accelerator on `code`, upsert 10k entities, run 1k trigram-served `attributeContains` queries, and the statistics show FILTERABLE at 10k/1k while SUBSTRING_ACCELERATED stays at 0/0 with null timestamps forever. The enum javadoc in `SchemaCapabilityUsageStatistics` (lines 311-313) documents precisely that reading as the signal to withdraw the accelerator.

evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
```diff
@@ -577,16 +641,425 @@ public ConsistencyReport getConsistencyReport() {
...
+	 *
+	 * Unregistering a consumer that is not the last one is unrestricted — the tree keeps its ids and nothing
+	 * structural happens.
+	 *
+	 * The same single-writer obligation as {@link #attachValueIdConsumer(String)} applies.
+	 *
+	 * @param consumerName the consumer's stable name
+	 * @see #attachValueIdConsumer(String)
+	 */
+	public void detachValueIdConsumer(@Nonnull String consumerName) {
```
novoj, line 713: **No production caller.** `detachValueIdConsumer` is reached only from `ValueIdTest`. `GlobalEntityIndex#reconcileTrigramIndexAbsence` (lines 683-685) removes the map entry only, so withdrawing the accelerator leaves the value-id column, the allocator, the per-write directory invalidation and the per-page id persistence on the shared value tree forever: every later insert mints an id, every flush writes the `int[]` column on every dirty leaf, every read after a write can trigger a synchronized O(V) directory rebuild, and `AttributeIndexLoader#restoreValueIds` re-attaches the allocator on every restart, all for zero consumers.

The premise text in this very method says whoever wires a consumer to an accelerator owes the drop path; it was never wired. Before wiring it, see the comment on `removeValueIdMinter`, which constrains how the detach can be done safely.

evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
```diff
@@ -577,16 +641,425 @@ public ConsistencyReport getConsistencyReport() {
...
 		final Comparable normalizedValue = (Comparable) this.normalizer.apply(value);
+		// the bucket count is the birth detector: the tree increments it exactly on the new-bucket branch, and it is
+		// transaction-aware, so a writer sees its own earlier writes and no one else's
+		final int bucketsBefore = sink == null ? 0 : this.buckets.size();
 		this.buckets.addRecord(normalizedValue, recordId);
+		if (sink != null && this.buckets.size() > bucketsBefore) {
+			notifyValueCreated(sink, normalizedValue);
+		}
 		this.dirty.setToTrue();
+		markValueIdDirectoryStale();
```
novoj, line 1093: **Stale flag raised for transactional writes too.** `markValueIdDirectoryStale()` (lines 960-962) is guarded on the allocator, not on `Transaction.isTransactionAvailable()`, so a transactional write whose data lands in a diff layer still flips the volatile on the shared live instance (here, at 1092 and at 1143).

During trunk incorporation of a transaction inserting N new distinct values, every read-only accelerated query arriving between two such writes enters the synchronized `refreshValueIdDirectory()`, walks every committed leaf and copies the location array (~8 MB for a 1M-value tree), and the next write raises the flag again: up to N redundant full rebuilds serialized on one lock, on the query path, while the merged index builds its own directory anyway. Suggest skipping the mark when a transaction is bound.

evita_engine/src/main/java/io/evitadb/index/bPlusTree/TransactionalBucketBPlusTree.java
```diff
@@ -852,7 +958,567 @@ public void bulkLoadPage(
...
+		final List<BPlusLeafTreeNode<K>> leaves = enumerateLeaves();
+		if (holdsAnyBucket(leaves)) {
+			Assert.isPremiseValid(
+				!Transaction.isTransactionAvailable(),
+				"Cannot drop the value id columns of a populated tree from inside a transaction — the walk would " +
+					"write through the base leaves and leak across transaction isolation."
+			);
+		}
+		this.valueIdMinter = null;
+		for (final BPlusLeafTreeNode<K> leaf : leaves) {
```
novoj, line 1142: **Latent today, but it constrains the fix for the missing detach path.** `removeValueIdMinter` nulls the minter and every leaf's `valueIds` in memory without dirtying any leaf. `InvertedIndex#isValueIdHighWaterDirty` then forces only the root out with `nextValueId = 0`, while untouched leaf pages on disk keep their `int[]` column.

On the next restart `AttributeIndexLoader` fails the premise at line 430 ("the root of filter index X and its leaf pages disagree about value ids"), or the sibling premise at 417 when some pages were rewritten, and the catalog does not open. Detaching must re-emit every live leaf page, or keep the column and drop only the consumer.

evita_engine/src/main/java/io/evitadb/index/trigram/TrigramSubstringSearch.java
```diff
@@ -0,0 +1,541 @@
...
+		@Nonnull TrigramIndex trigramIndex,
+		@Nonnull InvertedIndex sharedValueTree,
+		@Nonnull String rawPattern,
+		@Nonnull BiPredicate<String, String> exactPredicate,
+		@Nonnull LongUnaryOperator scannedDistinctValueCounter,
+		@Nonnull StringSearchShape shape
+	) {
+		// pre-flight rather than catch: the reverse lookup REFUSES inside a transaction, and the answer to that refusal
+		// is this fallback, so the condition is tested before anything commits to the accelerated path
+		if (Transaction.isTransactionAvailable()) {
```
novoj, line 335: **Accelerator silently off inside every read-write session body.** `EvitaSession#executeInTransactionIfPossible` (lines 2377-2383) binds a transaction for every `execute()` / `updateCatalog()` lambda on a catalog that supports transactions, so an import or sync job that queries `attributeContains` from the same read-write session it writes with never benefits from the accelerator it declared.

Not a wrong answer, but there is no signal, and neither `use/schema.md` nor `query/filtering/string.md` mentions the read-only / transaction caveat. Worth documenting at minimum; serving the committed directory for candidates whose leaves carry no diff layer would remove the limitation.

evita_engine/src/main/java/io/evitadb/core/query/filter/translator/attribute/AbstractAttributeStringSearchTranslator.java
```diff
@@ -169,6 +214,305 @@ private static AttributeBitmapFilter createAlternativeBitmapFilter(
...
+		// the accelerator is resolved BEFORE anything walks the target set, and the walk that prices the gate lives
+		// inside `TrigramSubstringSearch#match` - so an attribute that keeps no accelerator, by far the common case
+		// since only a SUBSTRING-declaring attribute has one, costs two map lookups per scope and nothing else
+		EnumMap<Scope, Formula> hoisted = null;
+		for (final Scope scope : processingScope.getScopes()) {
+			final GlobalEntityIndex globalEntityIndex = queryContext
+				.getGlobalEntityIndexIfExists(entityType, scope)
+				.orElse(null);
+			final TrigramIndex trigramIndex = globalEntityIndex == null ?
+				null : globalEntityIndex.getTrigramIndex(attributeIndexKey);
```
novoj, line 393: **Planning through a stale index after withdrawal.** The translator asks the index map and never the schema. Withdraw `SUBSTRING_SEARCH` from an attribute in LIVE and issue no further writes to it: the schema says plain filterable, the usage registry prunes the row, yet `getTrigramIndex(attributeIndexKey)` still answers and every `attributeContains` keeps being served by the trigram path while its heap stays resident indefinitely on a read-mostly catalog.

Results stay correct (the entry is dropped by the same write that could make it drift), but schema, statistics and plan disagree and memory is retained. Consulting the schema here, or dropping the entry eagerly on withdrawal, closes the gap.

novoj (commented):
Local Claude Code review (`/code-review 1483`), posted by hand because the CI review job on this PR keeps exiting green without a comment. Thirteen findings; one is a correctness bug that should block the merge. Details are in the inline comments.

**Blocking**
1. `InvertedIndex.java:1799` - the commit merge always takes the incremental directory rebuild and then clears the stale flag, so values written in place during warm-up after the last full build never enter the directory. Accelerated `attributeContains` / `attributeEndsWith` return false negatives for them until a restart.

**Should fix before merge**
2. `SchemaCapabilityUsageRegistry.java:626` - the `SUBSTRING_ACCELERATED` usage row is registered but never recorded on write or query; the statistics will tell operators to drop a hot accelerator.
3. `InvertedIndex.java:709` - `detachValueIdConsumer` has no caller; withdrawing the accelerator leaves the whole value-id machinery alive forever.
4. `TransactionalBucketBPlusTree.java:1112` - latent: `removeValueIdMinter` leaves on-disk leaf pages disagreeing with the root, so wiring (3) as it stands would make the catalog refuse to load. Constrains the fix.
5. `InvertedIndex.java:1062` - the stale flag is raised for transactional writes too, causing redundant synchronized O(V) directory rebuilds on the query path during trunk incorporation.

**Behaviour / documentation**
6. `TrigramSubstringSearch.java:328` - the accelerator is silently declined inside every read-write session lambda; undocumented.
7. `AbstractAttributeStringSearchTranslator.java:371` - a withdrawn accelerator keeps being served until the next write to that attribute.

**Defensive design**
8. `InvertedIndex.java:1511` - empty answer instead of a premise failure for an allocator-less tree with candidates.
9. `TrigramPostingAccumulator.java:258` - the bulk path accepts value id 0 where the incremental path throws.
10. `TrigramIndex.java:337` - the public `resolveCandidateValueIds(long[])` returns empty for a too-short pattern.

**Efficiency**
11. `InvertedIndex.java:1159` - a second tree descent to read back a just-minted id.
12. `TransactionalBucketBPlusTree.java:1387` - per-candidate premise and ThreadLocal lookups in the fused probe.

**Convention**
13. spike `TrigramCodec.java:82` - a second class of the same name beside the engine codec.

Checked and refuted, so not reported: NFD normalization mismatch; missing backfill when an accelerator is added later; value-id reuse; `_2026_2` Kryo reader drift; the `@Delegate` move dropping reference-attribute validation; proto field-number collisions; the `@Deprecated(since)` flip; the rewritten test assertions. No TODOs, no committed plans, ADR index current, javadoc complete on new files.

evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
```diff
@@ -828,6 +1420,190 @@ public Formula getRecordsMatchingFormula(@Nonnull Predicate<Serializable> valueP
...
+				"or drop the pattern."
+		);
+		Assert.isPremiseValid(
+			!Transaction.isTransactionAvailable(),
+			"Value ids cannot be verified while a transaction is open on this thread - the directory addresses the " +
+				"last published version of the tree while the leaves it reads are the transaction's own, so the " +
+				"verification would silently under-report. Take the scan fallback instead."
+		);
+		final LeafVersionAccumulator leafVersions = new LeafVersionAccumulator();
+		if (this.valueIdAllocator == null) {
```
novoj, line 1563: **Empty answer where the project rule says throw.** With non-empty candidates and no allocator this returns an empty `MatchedBuckets`. That is the *silently wrong* shape `AbstractAttributeStringSearchTranslator#resolveFromIndex` warns about in its own javadoc: it passes any test whose fixture is small enough for the empty answer to look plausible. Per the CLAUDE.md Defensive Design rule an unreachable state must throw, never no-op; a premise failure here is cheap and honest.

evita_engine/src/main/java/io/evitadb/index/trigram/TrigramPostingAccumulator.java
```diff
@@ -0,0 +1,344 @@
...
+	 * @param normalizedValue the normalized value the shared value tree holds
+	 * @param valueId         the id the tree allocated for it
+	 * @throws io.evitadb.exception.GenericEvitaInternalError when the value yields a trigram the counting walk
+	 * never saw, which means the tree changed between the two walks
+	 * @throws ArrayIndexOutOfBoundsException when a slot is offered more members than the counting walk counted for
+	 * it - its countdown reaches zero and the next store indexes -1. The over-fill carries no diagnosis of its own,
+	 * because the store below runs once per membership and the JVM already bounds-checks it; the mirror-image SHORT
+	 * fill IS diagnosed, in {@link #materialize()}
+	 */
+	private void collect(@Nonnull Serializable normalizedValue, int valueId) {
```
novoj, line 259: **Bulk path accepts value id 0, incremental path throws.** `collect` stores whatever id the leaf carries, so a reloaded page with a bucket at `UNASSIGNED_VALUE_ID` posts 0 into every trigram of that value, while `InvertedIndex#notifyValueCreated` (lines 1160-1164) rejects the identical input as a premise violation. Nothing can remove the phantom afterwards because `removeRecord`'s premise fires first on the dying-id side.

`locationOf` rejects id 0, so the cost is a permanent phantom candidate rather than a wrong answer, but the two paths should agree: assert here as the incremental twin does.

evita_engine/src/main/java/io/evitadb/index/trigram/TrigramIndex.java
```diff
@@ -0,0 +1,739 @@
...
+	 * produce stays the cheapest.
+	 *
+	 * @param trigrams the pattern's trigrams, as {@link TrigramCodec#extractUniqueTrigrams} produces them
+	 * @return the pattern's postings ordered by cardinality, or `null` when some trigram posts against nothing - and
+	 * likewise when there are no trigrams at all, which a caller must have refused before reaching here
+	 */
+	@Nullable
+	PatternPostings pricePattern(@Nonnull long[] trigrams) {
+		final int trigramCount = trigrams.length;
+		if (trigramCount == 0) {
```
novoj: **Null for an input the javadoc says must already have been refused.** `pricePattern()` answers *provably no value contains this* for a zero-trigram input, and the public `resolveCandidateValueIds(long[])` (lines 425-426) turns that into `EMPTY_INT_ARRAY`, i.e. *no match*, silently.

The only production caller, `TrigramSubstringSearch#match`, guards `trigrams.length == 0` first and the `long[]` overload is exercised only by tests, so this is a defensive-design gap rather than a live bug: assert here, or make the overload package-private.

evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
```diff
@@ -610,11 +1100,69 @@ public void addRecord(@Nonnull Serializable value, int... recordId) {
...
+	 * rests on. The insert path resolves the id after the fact because the id does not exist until the insert mints
+	 * it; the removal path, whose id instead stops existing when the bucket does, gets it back from the removal
+	 * itself and so pays not even the bucket counts — see
+	 * {@link #removeRecord(Serializable, ValueLifecycleSink, int...)}.
+	 *
+	 * @param sink            the sink to notify
+	 * @param normalizedValue the value the insert created a bucket for, already normalized
+	 */
+	private void notifyValueCreated(@Nonnull ValueLifecycleSink sink, @Nonnull Comparable normalizedValue) {
+		final int valueId = this.buckets.valueIdOf(normalizedValue);
```
novoj: **Second descent to read back a just-minted id.** `valueIdOf` performs a full root-to-leaf descent plus a leaf binary search (front-coded key decoding included) to resolve the id that `stampValueId(leaf, insertedAt)` had in hand at `TransactionalBucketBPlusTree` line 1544. A warm-up import of 1M distinct values pays 1M extra descents.

The removal path already returns the dying id from its own descent (`removeRecordReportingValueDeath`); `addRecord` could return the minted id (or `NO_NEW_BUCKET`) the same way.

evita_engine/src/main/java/io/evitadb/index/bPlusTree/TransactionalBucketBPlusTree.java
```diff
@@ -852,7 +958,567 @@ public void bulkLoadPage(
...
+		int valueId,
+		@Nullable Predicate<K> valuePredicate,
+		@Nullable byte[] containsPatternUtf8,
+		@Nonnull LongConsumer leafVersionSink
+	) {
+		Assert.isPremiseValid(!this.longPayload, "Int record-set API is not available on a long-payload tree!");
+		// the obligation `valueOf(int)` documents, stated as a premise here because this method - unlike that one -
+		// exists solely for the candidate-verifying consumer that must not silently under-report
+		Assert.isPremiseValid(
+			!Transaction.isTransactionAvailable(),
```
novoj, line 1417: **Per-candidate ThreadLocal work in the fused probe.** `InvertedIndex#getRecordsOfValueIdsMatching` already asserted once per query that no transaction is bound, yet this method re-asserts it per candidate and then performs three more `Transaction.getTransactionalMemoryLayerIfExists` lookups per candidate (`leaf.valueIdAt` via `getValueIds()`, `leaf.getKeyColumn()`, `leaf.getRecordsAt(slot)`).

For 50k candidates that is roughly 200k ThreadLocal reads inside the loop this probe exists to make cheap. The outer premise established that no layer exists, so the loop can read the base columns directly.

evita_test/evita_performance_tests/src/main/java/io/evitadb/spike/trigram/TrigramCodec.java
```diff
@@ -0,0 +1,399 @@
...
+ *
+ * # Distinct trigrams per value
+ *
+ * A value contributes each distinct trigram **once**: `"aaaaa"` yields a single `aaa` membership, not three.
+ * That is what makes the membership count `E` a function of distinct trigrams rather than of value length, and
+ * it is also what makes an update that does not change a value's trigram set a no-op for the index.
+ *
+ * @author Claude (P8 trigram-substring-index spike), FG Forrest a.s. (c) 2026
+ */
+public final class TrigramCodec {
```
novoj: **Second `TrigramCodec`.** This tracked `public final class TrigramCodec` sits beside `io.evitadb.index.trigram.TrigramCodec` and has already drifted from it. Force-tracking the spike is deliberate per the `.gitignore` comment, so this is only the name collision the CLAUDE.md naming rule forbids: renaming it (for example `SpikeTrigramCodec`) stops every stack trace and search hit from needing disambiguation.

evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
novoj: Fixed in 600b220ba.

This and the stale-flag finding below turned out to be one mechanism. `markValueIdDirectoryStale()` now raises the flag only when no transaction is bound: every node of the tree carries its transactional layer, so a write made inside a transaction lands in that layer and leaves the published leaves — the ones the directory addresses — byte-identical. The flag therefore says exactly *"a leaf changed content while keeping its instance token"*, which is the one condition under which the incremental rebuild is invalid, and the merge now reads it:

```java
if (this.valueIdDirectoryStale) {
    committedTree.rebuildValueIdDirectory();
} else {
    committedTree.rebuildValueIdDirectoryAfterMerge();
}
```

That is the "track in-place-since-last-build separately" shape rather than the cheap one — the narrowed flag *is* that tracking, so the full rebuild is paid only when a leaf really changed in place, and never for an ordinary transactional commit.

evita_engine/src/main/java/io/evitadb/index/usage/SchemaCapabilityUsageRegistry.java
novoj: Fixed in 382d92573, on both sides.

Write path: `reportAttributeTouched` now files the `SUBSTRING_ACCELERATED` row on exactly the test `SchemaCapabilityUsageRegistry#declaresCapability` mints it with — `getAcceleratorsInScope(scope).contains(SUBSTRING_SEARCH)`, unconditional on the filterable flag beside it, so the seeded row and the raised row cannot drift apart.

Query path: `AbstractAttributeStringSearchTranslator#hoistGlobalSubstringFormulas` records the request per scope. It records on the *declaration* rather than on the gate's verdict: a gate that declines the accelerator because the scan it displaces is cheaper is a planner cost decision, not an absence of demand, and counting only the accelerated plans would make a correctly-sized accelerator read as unused.

evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
novoj: Wired in c1dd3448a, and deliberately only as far as it can go.

`GlobalEntityIndex#reconcileTrigramIndexAbsence` now drops the value id consumer together with the index it drops, through a new `AttributeIndex#detachSharedValueIdConsumer`. So the registry no longer outlives the accelerator, and an **empty** tree outside a transaction gives its column back for real.

A **populated** tree keeps the column, and that is the finding below rather than an oversight: the drop dirties no leaf page while the root's high-water returns to unassigned, which is precisely the pairing `AttributeIndexLoader` refuses. Repairing it means re-emitting every live leaf page inside the ordinary entity write that observes the withdrawal — `O(V)` on one upsert, through the base leaves the tree refuses to touch inside a transaction. What is left behind is one `int` per distinct value on disk and a mint per newly created bucket, read by nobody, which is memory rather than correctness.

The drop is also skipped while a transaction is bound: the allocator and the minter are owner-resident, so clearing them writes straight through to the live index and an abort would restore the trigram index around a tree that had lost the ids its postings name.

The ADR's open item and its value-id decision row now say this instead of "detach has no production caller".

evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
novoj: Fixed in 600b220ba, together with the merge finding — `markValueIdDirectoryStale()` now returns without touching the volatile when a transaction is bound, so trunk incorporation raises nothing and the query path takes no synchronized `O(V)` rebuild between two of its writes.

The guard is sound rather than merely cheap: nodes are constructed with their transactional layer enabled, so a transactional write is confined to that layer and the directory keeps describing exactly the version every reader outside the transaction sees. Readers inside one are refused outright by `getValueById` and `getRecordsOfValueIdsMatching` anyway.

evita_engine/src/main/java/io/evitadb/index/bPlusTree/TransactionalBucketBPlusTree.java
novoj: Addressed in c1dd3448a, by making the unsafe shape unreachable rather than by making the walk safe.

`detachValueIdConsumer` now drops the minter only for an empty tree with no transaction bound, so `removeValueIdMinter` can no longer be reached with live pages behind it and the loader premise you name cannot be tripped from the detach path. The method's documentation now carries the persistence half of the restriction as well — the in-memory clear plus the high-water rewrite is the disagreement, and only re-emitting every live leaf page fixes it, which is not available on the write that observes a withdrawal.

evita_engine/src/main/java/io/evitadb/index/trigram/TrigramSubstringSearch.java
novoj: Documented in 044b3d295; the behaviour is unchanged on purpose.

`documentation/user/en/use/schema.md` now states that queries issued from inside a read-write session are always scanned, why (the accelerator answers from the last published version and cannot see the transaction's own writes), and what to do instead — read through a separate read-only session. Both `attributeContains` / `attributeEndsWith` notes in `query/filtering/string.md` link to it, and the pre-flight in `TrigramSubstringSearch#match` now names the user-visible consequence and the way out (serving the committed directory for candidates whose leaves carry no diff layer) instead of only the local reason.

evita_engine/src/main/java/io/evitadb/core/query/filter/translator/attribute/AbstractAttributeStringSearchTranslator.java
novoj: Fixed in 382d92573. The hoist now asks the schema before it asks the index map:

```java
if (!attributeDefinition.getAcceleratorsInScope(scope).contains(AttributeFilterAccelerator.SUBSTRING_SEARCH)) {
    continue;
}
```

A withdrawn accelerator therefore stops being planned through immediately rather than at the next write to the attribute, so schema, plan and statistics no longer disagree. What still survives until that write is the heap the map entry holds — the eager drop would have to happen in the schema mutation, which reaches the indexes with a transaction bound. `GlobalEntityIndex#getTrigramIndex` says so at the accessor.

It is also cheaper than what it replaces: one set-membership test per scope instead of a global-index lookup plus a map lookup.

evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
novoj: Fixed in 6b1c37306 — it is a premise now, with the empty answer kept only for the one honest case (no candidates at all):

```java
Assert.isPremiseValid(
    this.valueIdAllocator != null || candidateCount == 0,
    "Value ids cannot be verified against a tree that mints none - ..."
);
```

The javadoc on `AbstractAttributeStringSearchTranslator#resolveFromIndex` that warned about the crossing was updated with it: the `getRecordsOfValueIdsMatching` half is now caught, the `getValueById` half still answers `null` per id, so the crossing remains a mistake to avoid rather than one that is fully caught.

evita_engine/src/main/java/io/evitadb/index/trigram/TrigramPostingAccumulator.java
novoj: Fixed in 6b1c37306 — `collect` now asserts `valueId != ValueIdAllocator.UNASSIGNED_VALUE_ID` before storing anything, so the bulk path refuses exactly what `notifyValueCreated` refuses on the incremental one. One comparison per value per rebuild.

It is checked in `collect` rather than in `count` so the message can name the walk that would have stored the phantom; `count` would fail one walk earlier with nothing yet at stake.

evita_engine/src/main/java/io/evitadb/index/trigram/TrigramIndex.java
novoj: Fixed in 6b1c37306 — `pricePattern` refuses a trigram-less pattern instead of answering `null`. `null` is the strong answer *"provably no value contains this"*, and a pattern under three code points is the opposite case: the index holds nothing that could bound the search, so every value stays a possible match.

I kept the `long[]` overload public — it is reachable only from tests today, but package-private across a module boundary is fragile and the premise closes the hole regardless. Both javadocs now state the non-empty requirement. The one test that documents the guard as a calibration was updated: removing the guard now raises an internal error rather than silently producing the empty answer, which is the louder failure.

evita_engine/src/main/java/io/evitadb/index/invertedIndex/InvertedIndex.java
novoj: Fixed in 7f9414ebb. `TransactionalBucketBPlusTree#addRecordReportingValueBirth` is the mirror of `removeRecordReportingValueDeath`: `stampValueId` returns the id it minted, the insert hands it back, and `notifyValueCreated` takes it as a parameter instead of resolving it. `NO_CREATED_BUCKET` reports the join-an-existing-value case.

The birth detector changed with it — the tree's own new-bucket branch replaces the two bucket-count reads, so an insert that joins an existing value now costs nothing at all for the sink rather than two `size()` calls. A warm-up import of 1M distinct values loses 1M root-to-leaf descents plus their front-coded key decoding.

evita_engine/src/main/java/io/evitadb/index/bPlusTree/TransactionalBucketBPlusTree.java
novoj: Fixed in 7f9414ebb for the expensive half. Slot validation, the key test and the record read now live in `BPlusLeafTreeNode#recordsOfValidatedValueIdSlot`, which resolves the transactional layer **once** — a thread-local read plus a map probe, previously paid three times per candidate. The read order the tree method documents is preserved verbatim, and the leaf version token is a final field, so reading it after the probe is the same read.

I kept the per-candidate transaction premise. It is one thread-local read against three map probes removed, and the tree method is public API whose direct callers — the characterization tests among them — own that check themselves; the outer premise cannot speak for them.

evita_test/evita_performance_tests/src/main/java/io/evitadb/spike/trigram/TrigramCodec.java
novoj: Renamed in 3cc4e9ddf: the spike class is `SpikeTrigramCodec`, and the four spike files that resolved it by same-package lookup follow. `TrigramArmSweep`, which explicitly imports the engine's `io.evitadb.index.trigram.TrigramCodec` and therefore silently used that one all along, is unchanged and now unambiguous.

Worth noting for anyone repeating this: the rename had to be staged with `git mv` rather than a plain move, because the spike folder is git-ignored and its tracked files were force-added — a plain rename would have silently untracked the class. The package README says why the name differs from the engine's.

Issue #1490, opened 2026-09-03
Transaction conflicts are counted as internal engine errors


novoj:
## Problem

`ConflictingCatalogMutationException` is counted as a genuine engine fault. Every construction
increments `io_evitadb_errors_total{error_type="ConflictingCatalogMutationException"}`, and the same
counter (`ObservabilityManager.EVITA_ERRORS`) is what `ObservabilityProbesDetector.checkEvitaErrors`
reads, so each conflict also lights the `EVITA_DB_INTERNAL_ERRORS` tile.

A transaction conflict is not a fault. It is an expected, recoverable outcome of concurrent writers,
and the engine already says so in one place — `AbstractTransactionStage.handleException` deliberately
special-cases it to `debug` instead of `error`. It also already has a dedicated, better-labelled
counter: `io_evitadb_transaction_transaction_conflict_total`, carrying `conflictPolicy`,
`resolutionLayer` and `conflictScope` (added in `dda2383ce`, shipped since v2026.2.4).

That earlier change added the parallel metric but never opted the exception out of error monitoring,
so conflicts are now counted twice — once correctly, once as an internal error.

## Observed

On a production deployment (4 catalogs), over two days:

| | conflicts | `errors_total` |
|---|---|---|
| v2026.2.6 | 6 | 6 |
| v2026.2.5 | 8 | 16 |

(The 2:1 on v2026.2.5 is the separate double-count already fixed by `7bbef87f4` in v2026.2.6 — not
in scope here.)

## Fix

Add `@NotMonitored` to `ConflictingCatalogMutationException` **and** to
`ConflictingCatalogCommutativeMutationException`. Both are needed: the marker is not `@Inherited`
and `ObservabilityManager.MONITORED_ERROR_TYPES` evaluates it against the concrete runtime class,
so the parent's annotation does not cover the subclass.

The gate sits ahead of the metric, the `EVITA_ERRORS` counter and `ErrorOriginLogger`
(`ObservabilityManager:218`), so one marker removes the conflict from the metric, the health tile and
the origin log together.

**Accepted cost:** these two types lose `ErrorOriginLogger` stack-trace coverage. Acceptable, because
the conflict path already produces both a dedicated metric with finer labels and a log line carrying
the conflict key, catalog version and resolved policy.

## Follow-ups (deliberately not in this fix)

1. **gRPC status code.** `GlobalExceptionHandlerInterceptor:120` maps a conflict to `INTERNAL` (13)
[...]

PR #1491, opened 2026-09-03
fix: stop counting transaction conflicts as internal engine errors

novoj:
Ref: #1490

Hotfix twin of the `dev` PR. The issue is closed by the `dev` PR so the fix is not marked done until it is on the mainline.

## What

Adds `@NotMonitored` to `ConflictingCatalogMutationException` and to
`ConflictingCatalogCommutativeMutationException`.

## Why

A transaction conflict was being counted as a genuine engine fault. Every construction moved
`io_evitadb_errors_total{error_type="ConflictingCatalogMutationException"}`, and
`ObservabilityProbesDetector.checkEvitaErrors` reads the very same counter
(`ObservabilityManager.EVITA_ERRORS`), so each conflict also raised the `EVITA_DB_INTERNAL_ERRORS`
signal.

A conflict is an expected, recoverable outcome of concurrent writers. The engine already says so in
one place — `AbstractTransactionStage.handleException` deliberately special-cases it to `debug`
instead of `error` — and it already has a dedicated, better-labelled counter,
`io_evitadb_transaction_transaction_conflict_total`, carrying `conflictPolicy`, `resolutionLayer`
and `conflictScope`. That metric shipped in v2026.2.4 but never opted the exception out of error
monitoring, so conflicts were counted twice: once correctly, once as an internal error.

## Why both classes

`NotMonitored` is not `@Inherited`, and `ObservabilityManager.MONITORED_ERROR_TYPES` evaluates it
against the concrete runtime class. Annotating only the parent would leave commutative conflicts
counted under their own label.

## Accepted cost

Both types lose `ErrorOriginLogger` stack-trace coverage, since the same gate at
`ObservabilityManager:218` sits ahead of the metric, the counter and the origin logger. Acceptable:
the conflict path already emits a dedicated metric with finer labels plus a log line carrying the
conflict key, catalog version and resolved policy.

## Verification

- `mvn -pl evita_api -am compile` — BUILD SUCCESS.
- `javap -v` confirms both classes carry `NotMonitored` under `RuntimeVisibleAnnotations`, which is
  what the runtime `isAnnotationPresent` gate requires (`CLASS` retention would compile and then
  silently opt nothing out).

## Not in this PR

Two findings surfaced while diagnosing this; both are written up on #1490 and deliberately left out
of a one-line metric fix:

1. `GlobalExceptionHandlerInterceptor:120` maps a conflict to gRPC `INTERNAL` (13) and logs it at
   `ERROR`. The canonical code is `ABORTED` (10), which client retry policies treat as retryable.
2. `TransactionConflictEvent`'s javadoc claims conflicts are a subset of
   `TransactionFinished(resolution=ROLLBACK)`. They are disjoint — a conflicted transaction reports
   `COMMIT`, because `SessionRegistry.reportTransactionResolution` reads session-side
   `isRollbackOnly()` and fires before the conflict stage runs.

### DIFF
--- a/evita_api/src/main/java/io/evitadb/api/exception/ConflictingCatalogCommutativeMutationException.java
+++ b/evita_api/src/main/java/io/evitadb/api/exception/ConflictingCatalogCommutativeMutationException.java
@@ -24,6 +24,7 @@
 package io.evitadb.api.exception;
 
 import io.evitadb.api.requestResponse.mutation.conflict.ConflictKey;
+import io.evitadb.exception.NotMonitored;
 
 import javax.annotation.Nonnull;
 import java.io.Serial;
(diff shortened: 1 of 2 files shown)


novoj:
Mainline twin: #1492 (targets `dev`). Both carry the identical change; this one ships it on the release line.

novoj:
Merged into `hotfix-2026-2-7` rather than directly into `master`: this change ships as part of the aggregated 2026.2.7 hotfix in #1517, which is the single release-triggering merge for the three fixes queued for that patch. Review history here stays intact — the branch was merged with `--no-ff` and no rebase, squash or content change.

Issue #1501, opened 2026-09-07
EQUALIZED histogram: bucketing collapses on tied prices and relativeFrequency is an unstable estimate


novoj:
`HistogramBehavior.EQUALIZED` shipped in 2026.2 with two independent defects. Both are visible on production data today.

Full derivation, measurements and rejected-alternatives history will land as an ADR under `documentation/adr/` with the implementing PR.

## Defect 1 — `relativeFrequency` is an unstable estimate

The height is `occurrences / bucketWidth`, sum-normalised. The quantity is legitimate — on an equal-pixel equalized axis it is the density-quantile function — but it is estimated from **a single gap between two adjacent prices**, so one reprice moves it by orders of magnitude.

On a production Senesi category (`zrcadla-a-galerky`, 3 237 products):

- the bucket holding **81 products** renders at `relativeFrequency` **35.15**
- the bucket holding **338 products** renders at **6.03**

A bucket with a quarter of the products draws almost six times taller. On the evitaDB test shop the same defect inverts a 5-product mass against a single product (0.99 vs 99.01), and the shipped `histogram.md` example shows it mid-histogram (3 products ten times taller than 4).

## Defect 2 — bucketing collapses on tied prices

`EqualizedHistogramDataCruncher.java:271-299` advances **one** bucket per iteration however many quantile targets a heavy value crossed (`if` where a `while` is needed). A distinct value's weight being atomic is inherent and fine; failing to charge it for the targets it consumed is not.

On a charm-priced catalogue asking for 20 buckets of 100: **13 of 20 buckets under half their target**, 32% of the catalogue in bar one. Tie-free data is unaffected — it degenerates only on ties, which is exactly how retail pricing looks. Confirmed in production: the Senesi category holds 3 237 products across only **135 distinct prices**, with 241 at 999 Kč, 176 at 799, 172 at 899.

## Proposed fix

**Bucketing** — thresholds are the empirical inverse CDF sampled at `k/B`, de-duplicated, compared in integer arithmetic. A price absorbing ≥2 quantile targets also emits a threshold at the next distinct price, closing the mass into its own bucket; the budget proof keeps the result ≤ `bucketCount`. Point-mass plateaus collapse to one selectable position, so every returned threshold is a distinct price and every slider position yields a different result set.

[...]

PR #1502, opened 2026-09-07
docs: rewrite faceted filter blog post for a wider audience

novoj:
Rewrites `documentation/blog/26-faceted-filter-choreography.draft` for a developer / e-commerce-owner audience rather than for evitaDB users. Still a draft (`draft: true`), so nothing is published by merging this.

## What changed

**Removed** — the constraint-level walkthrough: the three carrier families, the four-baselines matrix, the `histogramHaving` vs `attributeBetween` section, the filter-side/require-side pairing table, the evitaLab tour, the demo-schema aside and the query example. Zero evitaDB API identifiers remain in the file.

**Kept** — the mechanics and caveats, which are stack-independent: the dead-end problem, the four shopper invariants, why a slider must not collapse under its own handle, why a facet needs two numbers rather than one, and the existing sourced UX research.

**Expanded** — the price slider is now the centre of the article (~45% of the body): histogram equalization and its origin in image processing, the trap that equalizing the pivots leaves the bars meaningless, why charm pricing turns a catalogue into point masses, and the plateau collapse as a case of a clean construction yielding to a product requirement. Names the scientific background the argument rests on — the quantile function, kernel density estimation, bandwidth selection, and the normal-theory assumptions retail price data violates.

## Accuracy

Every figure is measured rather than illustrative. The production numbers (3 237 products across 135 distinct prices; 241 at 999 Kč) come from a live catalogue captured over JDWP; the density failures come from the same investigation. The post makes **no claim that the reworked density calculation ships today** — the related work is #1501 and is still open.

## Two decisions for the reviewer

- `date: '14.05.2026'` is unchanged, but the post now references work done after 2026.2 shipped (28.07.2026). Probably wants updating before publication.
- One forward-looking sentence says the rebuild is under way and the reasoning will be published with it. It deliberately carries no issue number, since that is an internal detail for this audience. Happy to drop the sentence entirely if the post should make no commitment at all.

### DIFF
--- a/documentation/blog/26-faceted-filter-choreography.draft
+++ b/documentation/blog/26-faceted-filter-choreography.draft
@@ -1,7 +1,7 @@
 ---
-title: The hidden choreography of a faceted filter panel
+title: The hidden mathematics of a faceted filter panel
 perex: |
-  Every well-built e-shop filter panel runs on a delicate choreography. The shopper drags a slider, ticks a checkbox, and the panel must redraw without ever feeling broken — collapsing sliders, facets that read zero, price ranges that shrink to a sliver are the trust-killers. The unspoken UX rules behind those panels translate to a surprisingly precise set of database semantics. In evitaDB they take the shape of `userFilter` and three carrier families. This post walks through that choreography from the shopper's screen to the query.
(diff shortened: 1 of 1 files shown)


novoj:
Superseded by #1506, which carries the same commit against `dev` (the correct base — this branch was cut from `master`, so retargeting would have dragged in the whole 2026.2.6 hotfix chain). Closing here.

PR #1576, opened 2026-09-14
fix: release the resources a closed engine still holds

novoj:
Three independent defects, all of the same shape: `close()` released most of what it owned, but not all
of it, and what it left behind was held by something that outlives the instance.

Found while tracking down why the functional suite exhausts its surefire fork heap. The suite was the
symptom; two of the three are engine defects that also affect a running server.

### 1. The engine persistence service never closed its output keeper

`DefaultEnginePersistenceService#close()` released the mutation log, the off-heap memory manager and the
folder lock, but not the `ObservableOutputKeeper` it builds for itself. The keeper holds every output
buffer the service ever borrowed, each sized at `StorageOptions#outputBufferSize()` - 2 MB by default -
and its only other release path is its own inactivity task, which runs on a scheduler a closing engine
has already shut down.

The catalog-side service already does this correctly, which is the shape of the evidence: 1,443 engine
services live against ~123 catalog services at one instant, and 1,443 x 2 MB accounts for 2.9 GB of the
3.4 GB of live byte arrays.

### 2. Flight Recorder periodic hooks were never handed back

`FlightRecorder.addPeriodicEvent(Class, Runnable)` appends to `jdk.jfr.internal.RequestEngine#entries`, a
static list that lives as long as the JVM; the only way out is `removePeriodicEvent`, matching on object
identity. A static of a JDK class is a GC root. Four sites registered hooks and none released them, and
the per-catalog hook was an inner class - so it held the engine, and through it every catalog,
persistence service and output buffer the engine reached.

The existing self-removal could not help: periodic hooks only run while a recording is active, and the
removal sat inside `if (Evita.this.isActive())` - gating the one path that could unregister a hook on the
instance being alive, when the instances needing unregistration are exactly the dead ones.

**This half is a production defect independent of the test suite.** `emitCatalogStatistics` runs on every
`ModifyCatalogSchemaMutation`, not only on catalog creation, so a long-lived server registered another
permanent hook on every schema change - emitting that catalog's statistics once per change it had ever
seen, and lengthening the list every JFR period walks.

### 3. The test scheduler's shutdown did not terminate it

`ImmediateScheduledThreadPoolExecutor` kept delayed tasks scheduled across `shutdown()`, and two of its
overrides reported outcomes that were not true - `awaitTermination` returned `true` unconditionally,
which made `Evita#shutdownScheduler` skip its `shutdownNow()` fallback, and `isTerminated` answered from
a local flag. On its own this recovers no memory; it is here on its own terms.

## Measured

Functional suite, parallelism 4, 8g fork - the CI conditions, reproduced locally.

| | before | + keeper close | + JFR fix |
|---|---|---|---|
| outcome | **OutOfMemoryError** | no OOM | no OOM |
| tests | 23,079 | 23,975 | 23,976 |
| failures | 7 failures + 7 errors | 0 failures, 1 error | 0 failures, 1 error |
| wall | 09:43 | 06:36 | 05:08 |

The remaining error is the Docker-dependent `ExportS3ServiceTest`, expected off a Docker host.

Live objects after a forced full GC, all at the same point in the test fork's life:

| | before | + keeper close | + JFR fix |
|---|---|---|---|
| `RequestEngine$RequestHook` | 7,389 | 6,687 | **191** (JVM baseline 47) |
| live `Evita` instances | 1,273 | 1,151 | **153** |
| live byte arrays | 2,982 MB | 662 MB | - |
| total live heap | 3,523 MB | 1,083 MB | - |

Before, the hook registry only climbed - 47, 7,389, 9,058, 9,097 within one run, with no downward path.
After, growth decays towards the number of engines actually alive: 86, then 27, then 13 per 40 seconds.

## What this does not fix

Roughly 105 fully-constructed engines are still reachable at 122 s of fork life, against a harness peak
of 24 simultaneous - down from ~1,250, no longer a heap risk, retainer unnamed. Separately, ~52 `Evita`
objects that never reached `Evita.java:608` (constructions that threw partway) are also retained; they
are nearly free, and they were invisible under the JFR husks.

The dedup half of (2) has no regression test. The obvious assertion - the hook map's size - would not
kill the obvious mutation, since `put` instead of `putIfAbsent` keeps the map at one entry while
registering N hooks. The faithful version is a real `jdk.jfr.Recording` with `period=everyChunk`,
counting `io.evitadb.storage.CatalogFlush` rows per catalog name; deliberately left out rather than ship
a test that cannot fail for its own reason.

## Note on the CI fork heap

#1574 raised the CI fork heap to 9g because the suite exhausted 8g; #1575 has since returned it to 8g,
recording why 9g failed. So `dev` is back on the budget the suite was overrunning, and these fixes
address the cause rather than the budget: at parallelism 4 with an 8g fork the suite now completes with
the collector reclaiming, where before it died inside its own heap.


novoj:
Thanks — this was a useful review. All four points addressed; one of them turned out to be more serious than either of us had it.

**MAJOR, missing ADR — accepted, and you changed my mind.** I had decided against a record on the grounds that the diff plus commit messages carried it. The argument that moved me is the one about scope: the next person who calls `FlightRecorder.addPeriodicEvent` anywhere in the engine walks into the identical trap, and a commit message on three files in `evita_engine` is not findable from another module. That is the "reasoning outlives the change" limb, and it is genuinely cleared. Added in `33f89c668` as `documentation/adr/2026-09-14-closed-engine-resource-release.md`, point-decision shape, with three real forks and their rejected alternatives — the registration ordering, owner-scoped release versus the hook retiring itself, and the test strategy. Index regenerated in the same commit.

**MINOR 2, `FormulaCacheVisitorTest` — accepted, and it was wider than reported.** `CacheAnteroomTest` does the same thing. Both now close the `CacheAnteroom` that owns the eden in `@AfterEach`, in `5999c82ed`. This also explained a measurement I had never accounted for: every heap histogram taken during the investigation showed exactly five live `CacheEden` instances, before and after the fix. `FormulaCacheVisitorTest` has exactly five test methods.

**NIT, the TOCTOU window — not documented as residual; fixed, because it had a companion that made it reachable.** A local bug-hunting pass found that `registerEngineStatisticsHook` recorded the hook *before* registering it — the reverse of the order `emitCatalogStatistics` uses and of the rule its own comment states. A concurrent `close()` could read the hook out of the list, call `removePeriodicEvent` on a hook not yet registered (a no-op), clear the list, and strand the registration that followed. That is precisely the failure the tracking exists to prevent, and it is reachable: `MetricHandler#registerHandlers` waits one minute on `metricTask.getInitialized()`, and on timeout it returns and logs while the callback that calls `emitStartObservabilityEvents` stays pending, free to fire during close.

Reordering alone is insufficient, since `retireStatisticsHooks()` reads the list too. Both methods now re-read `active` after recording; the two orders interlock, so at least one side observes the other and exactly one retires the hook. Fixed in `f616534c3`.

The same pass also found that this PR's own test overstated its coverage. `EvitaJfrHookRetentionTest` claimed to walk every registration path, but reachability only catches a retainer that reaches the watched object, and `CacheEden` holds nothing leading back to an engine — disabling `CacheEden#close()`'s `removePeriodicEvent` left the test green. The cache now has its own reachability test that the counterfactual does fail, and the javadoc says which sites each test can see.

**MINOR 1, no issue reference — declined, with the substance addressed.** The `ci-raise-surefire-heap-to-9g` precedent shows issue-less branches with a substantial commit body are accepted here, and opening a public issue on defects felt like the maintainer's call rather than mine. Your real point was that the follow-ups need a home, and they now have one: the ADR's "Consequences & open follow-ups" carries the still-reachable engines with the instrument to use, the partially constructed ones, the untested deduplication path, and one further defect the pass turned up — a constructor that throws can never be closed, because `active` is set at the end of the constructor and `close()` is `compareAndSet(true, false)`, so everything acquired from `SessionKiller` onward leaks with no recovery path at all. Deliberately not fixed here: it is wider than hook lifecycle, needs a test that fails a construction on purpose, and a hooks-only guard would read as handled while the folder lock still leaked.

PR #1464, opened 2026-08-30
Remove customer attribution from the repository

novoj:
`documentation/adr/` and the rest of `documentation/` publish with the source, so a customer's
catalog name, the shape of their pricing, or the fact that they hit a corruption bug is not ours to
publish. This removes every client name from the repository and adds the rule that keeps it out.

**Every measurement is preserved.** Corpus sizes, cardinalities, percentiles, latencies, speedups,
ratios — none of it identifies anyone, and stripping it would gut the records. Only the attribution
is gone. The diff is almost entirely one noun swapped for another.

## What was found

Four client names, not the two we started looking for. `cnc` turned out to be a non-issue on `dev` —
all 22 hits were PNG/TTF/WOFF binaries and one base64 blob.

| token | where | replaced with |
|---|---|---|
| `senesi` | 344 hits / 79 files | the production catalog / the price-heavy catalog |
| `keramikaSoukup` | 21-file Java package + docs | the single-price catalog |
| `signal` | 21-file Java package + docs | the variant-heavy catalog |
| `decodoma` | ~30 hits, ADR design docs + spike code | the gate catalog |

Datasets are now named for the workload shape they exercise, which is what the surrounding text is
actually about. Where a record genuinely compares two of them — the PriceSuper analysis weighs a
price-shallow catalog against a price-dense one — both keep a shape label so the comparison still
reads. The research assignment's dataset-selection tables named five candidate shops; they are now
Catalog A–E.

`signal` needed care: it is also ordinary English throughout the engine (`GrpcCatalogState`,
`CatalogState`, `ReevaluateExpressionExecutor`). Only the `performance/signal/` package and explicit
`Signal.cz` / `Signal-nabytek.cz` mentions were touched — never a blanket replace.

## Renames a reviewer should check

**Java packages** (`evita_test/evita_performance_tests`), 63 files:

| old | new |
|---|---|
| `io.evitadb.performance.senesi` / `Senesi*` | `io.evitadb.performance.priceHeavyCatalog` / `PriceHeavyCatalog*` |
| `io.evitadb.performance.keramikaSoukup` / `KeramikaSoukup*` | `io.evitadb.performance.singlePriceCatalog` / `SinglePriceCatalog*` |
| `io.evitadb.performance.signal` / `Signal*` | `io.evitadb.performance.variantHeavyCatalog` / `VariantHeavyCatalog*` |

**Files** (each keeps its date prefix, so it still sorts with its record):

| old | new |
|---|---|
[...]

PR #1598, opened 2026-09-16
refactor: rename customer-named benchmark packages and their references

novoj:
Ref #1454. Reworked from the stale `1454-remove-customer-attribution` branch (PR #1464, closed), rebuilt on current `dev`.

### What was still violating

`CLAUDE.md` marks code identifiers, comments and hardcoded paths as a surface where a customer name is **never** acceptable. PR #1465 already scrubbed records, documentation and code comments, but the benchmark fixtures under `evita_test/evita_performance_tests` kept three production customers' names in their package names, their class names and the catalog name each data source returns.

### The change

Three packages and the classes beneath them are renamed for the property each dataset exercises rather than for its origin — `priceHeavyCatalog`, `singlePriceCatalog`, `variantHeavyCatalog` — 63 files in all. The values returned by `getCatalogName()` follow, so a run resolves its dataset directory by the same neutral name.

Three further sites are cleared, all added after the earlier pass and therefore outside its reach:

- `IsolatedWarmupLoadBenchmark` — the gRPC message-size javadoc, and the `CATALOG_NAME_PROPERTY` default
- `documentation/adr/2026-09-10-jdk21-virtual-threads-and-scoped-values/evidence/04-preview-and-docs.md` — one benchmark name in the fork-args table

### Deliberately untouched

The surfaces `CLAUDE.md` exempts — `documentation/performance/**`, blog, publishing and research — keep naming their datasets. That exemption is explicit in the rules table ("consent covers these"), so scrubbing them would be a regression, not a fix. The original branch did scrub them, which is one reason it is not simply rebased here. Its policy commit is dropped too: that text is already in `dev`.

### Verification

`mvn test-compile -P full` on JDK 21 — **BUILD SUCCESS**, with `evita_performance_tests` compiled (module 31/33), so every renamed file was exercised. Grepping the three retired names across the whole repository outside the exempt surfaces returns **zero hits**.

### Follow-up for whoever merges this

Benchmark dataset directories on developer machines still carry the old names and need renaming to the three neutral ones before the next run.


novoj:
Superseded: the branch name and the commit's `Ref:` line both borrowed #1454, which is an unrelated issue (Trigram substring index). Reopened as a branch with no issue reference, following PR #1465's precedent.

Issue #1545, opened 2026-09-10
Price termination formulas: struct-of-arrays columns, merge-join with SIMD galloping probes and a vector range predicate


novoj:
## Summary

Part of #1539. The three price termination formulas cannot be vectorized as written: they walk entity ids from the
delegate bitmap, call a per-entity callback over `PriceRecordContract` objects found by a binary search over a
shrinking window, deduplicate inner records through a hash map per entity, and branch per record. A struct-of-arrays
layout with a sorted merge-join is faster on its own and opens the probes and the predicate to SIMD.

## Where

- `LowestPriceTerminationFormula.computeInternal` (`:453-590`), `SumPriceTerminationFormula.computeInternal`
  (`:331-440`, allocates an `IntObjectHashMap` and a `CumulatedVirtualPriceRecord` per entity),
  `PlainPriceTerminationFormulaWithPriceFilter.computeInternal` (`:243-320`)
- `ResolvedFilteredPriceRecords.PriceRecordIterator` (`:110-215`): `Arrays.binarySearch` over an `int[] entityIds`
  side array inside a narrowing window, then a callback per record object
- `PricePredicate` (`:113-116`): an `int` range on `priceWithTax` or `priceWithoutTax`
- `PriceRecord` (`:51-56`): a record of five `int`s, array-of-structs behind an interface with three implementations

## The rewrite (ADR §6.2–6.3)

1. **Columns.** Each `FilteredPriceRecords` carries parallel `int[]` columns sorted by `(entityId, innerRecordId)`:
   `entityId`, `innerRecordId`, `priceWithTax`, `priceWithoutTax`, `internalPriceId`. Per-query variant first (a
   one-pass copy of the arrays that already exist, `SortingForm.ENTITY_PK`); a persistent column layout of the price
   pages in `PriceListAndCurrencyPriceSuperIndex` is a later, separate decision (storage format, Kryo,
   backward compatibility).
2. **P-way merge instead of P binary searches per entity.** One cursor per price list; for each entity of the sorted
   batch advance every cursor to its first row; the per-list runs for that entity are sorted by inner record id, so
   the highest-priority list wins per inner record by walking the runs together — no hash map, no callback, no
   per-entity allocation.
3. **Galloping for the sparse case, with vector probes.** When the delegate is much sparser than the rows, the cursor
   advance gallops; the probe is one `IntVector.compare(GE, e)` over 8 (AVX2) or 16 (AVX-512) consecutive ids plus
[...]

Issue #1540, opened 2026-09-10
Facet counts and impacts: evaluate cardinality without materializing the intersection (prerequisite for the SIMD kernels)


novoj:
## Summary

Prerequisite for the SIMD work (part of #1539): every facet count and impact is obtained by *materializing* the
intersection and reading its size, although `PersistentRoaringBitmap.andCardinality` and
`FastAggregation.andCardinality` exist. This is a scalar change that removes at least half of the arithmetic and all
of the allocation on the facet path, and it changes the baseline any vector kernel would be measured against — so it
comes first.

## Where

- `MemoizingFacetCalculator.calculateImpact` (`:110-127`): `hypotheticalFormula.compute().size()`
- `ReferenceSummaryProducer` (`:1312-1318`): `this.resultFormula.compute().size()`
- `AbstractFacetFormulaGenerator` (`:582`), `ImpactFormulaGenerator.hasSenseAlone` (`:198`, `:207`):
  `compute().isEmpty()`
- `compute()` of an `AndFormula` → `RoaringBitmapBackedBitmap.and` (`:297-300`) → `PersistentRoaringBitmap.and`
  (`PersistentRoaringBitmap.java:286-311`) → `Container.and`; `BitmapContainer.and` (`:221-235`) runs
  `andCardinality` (pass 1) and then a second 1024-word pass or `Util.fillArrayAND` into a fresh container; then a
  `RoaringArray` and a `PersistentRoaringBitmap` are allocated; `getCardinality()` sums cached container counts.

## Proposed change

A `computeCardinality()` on `Formula` with `compute().size()` as the default, implemented natively by `AndFormula`,
`FacetGroupAndFormula`, `FacetGroupOrFormula` and the constant/bitmap formulas by delegating to
`PersistentRoaringBitmap.andCardinality` / `FastAggregation.andCardinality` (and the `or` counterparts). The three
consumers above call it. Memoization semantics stay: a formula that was materialized answers from the cached bitmap.

A second, optional step once this exists: evaluate the N facet bitmaps of one reference against the *same* base
container in a loop that keeps the base in L1 (base-outer, facets-inner), which is the shape the later vector
kernels like best.

## Gate

`ArtificialFacetFilteringAndSummarizingCountState`, `ArtificialFacetAndHierarchyFilteringAndSummarizingImpactState`
and their `Client*` twins in `evita_performance_tests` on a production-shaped catalog: allocation per facet near
[...]

Issue #1539, opened 2026-09-10
Adopt the JDK Vector API (SIMD) for evitaDB's internal kernels


novoj:
## Summary

Umbrella for adopting the JDK Vector API (`jdk.incubator.vector`, SIMD) inside evitaDB's own kernels. The HNSW work
planned for this release (fulltext ADR `2026-08-24`, P6 spike) is expected to bring the incubator module into the
runtime anyway; this line of work asks where the same module pays off internally and how it ships. The feasibility
analysis with the per-candidate verdicts is recorded in the ADR
`documentation/adr/2026-09-10-simd-vector-api-feasibility/` (analysis in `analysis.md`).

## What the analysis found

- evitaDB's query arithmetic is set algebra on 16-bit chunks (the vendored roaring containers) plus a small amount of
  per-record filtering. SIMD fits the container kernels that combine a bitwise op with a *reduction* (population
  count) and sorted-set intersection when both sides have similar size. It does not fit anything that walks objects,
  chases pointers or branches per element.
- The facet generators (`FacetFormulaGenerator`, `ImpactFormulaGenerator`) only build formula trees; the arithmetic is
  in the containers. Before any kernel, facet counts should stop materializing intersections (they use
  `compute().size()` where `andCardinality` exists).
- The price termination formulas cannot be vectorized as written; a struct-of-arrays layout with a merge-join makes
  them faster on its own and opens the predicate and the probes to SIMD.
- The range index's `JoinFormula` + `DisentangleFormula` pair is a signed multiplicity count; a per-chunk counting
  kernel replaces both and emits result words straight from vector masks — the cleanest SIMD fit found.
- On the installed HotSpot 21.0.12 (verified in `libjvm.so`): lane-wise `BIT_COUNT` is intrinsified on every AVX2
  CPU (lookup-table path) and natively on AVX-512 VPOPCNTDQ; `compress`/`expand` are AVX-512 only; 16-bit `rearrange`
  is intrinsified on AVX2 at 128 and 256 bits.

## How it ships

Optional module (`requires static jdk.incubator.vector`), a provider that self-detects the module, C2 and the CPU at
class-init and otherwise uses the scalar twin, `--add-modules jdk.incubator.vector` in the launchers and test
arguments, differential tests scalar-versus-vector. Forgetting the flag means the scalar path runs, not a crash. The
Java driver is unaffected (`evita_roaring_bitmap` and `evita_engine` are not on its dependency path).

## Child issues, in the recommended order

[...]

PR #1608, opened 2026-09-19
perf: optional Vector API provider, fused word kernels and a CRoaring-style lazy array union for the roaring containers

### LINKED ISSUE #1541: Vector API provider: optional `jdk.incubator.vector` wiring with a scalar fallback, CPU gating and differential tests
## Summary

Infrastructure for every vector kernel (part of #1539): make `jdk.incubator.vector` an *optional* dependency with a
scalar fallback, so that a missing module, a non-C2 JIT or an unsuitable CPU runs the existing code and never a slow
Java lane loop.

## Design (from the ADR, §3)

1. `evita_roaring_bitmap/src/main/java/module-info.java` (and `evita.engine` where a kernel lives) declares
   `requires static jdk.incubator.vector;`. Neither module is reachable from the Java driver, so the JDK 17 floor and
   its enforcer rule are untouched.
2. `maven-compiler-plugin` `<compilerArgs>` for those modules gains `--add-modules jdk.incubator.vector` (root POM today
   passes only `-parameters`, `pom.xml:750-752`). javac warns about the incubating module; the build does not fail
   on warnings.
3. A provider holder (name after the naming check; `VectorKernels` is free today) resolves once at class-init: module
   present in `ModuleLayer.boot()`, no kill-switch system property, `VectorShape.preferredShape().vectorBitSize() >=
   256` (128-bit NEON allowed, gains little), HotSpot with C2 enabled; then the Panama implementation is loaded
   reflectively, otherwise the scalar one. Logged once at INFO and exposed through the system statistics.
4. `--add-modules jdk.incubator.vector` in `docker/entrypoint.sh`, `evita_server/run-server.sh`,
   `evita_server/dist/run.sh`, the surefire `argLine` (`pom.xml:809-815`), the JMH fork arguments and the IDE run
   configurations; documented in `operate/run.md` and the Java-embedding page. Forgetting it means the scalar path
   runs.
5. Tests: scalar-versus-vector differential tests on randomized inputs; the roaring suite with the vector provider
   forced on and forced off; CI runs both modes. Test mode forces the scalar provider unless a test asks otherwise.
6. Per-kernel enable switches, because the gain differs by CPU generation (native `vpopcntq` versus the lookup-table
   popcount; `compress` AVX-512 only).

## Verified facts this relies on

Incubator modules do not mark class files (no minor version 65535), unlike `--enable-preview`; the API used
(`fromArray`/`fromCharArray`, `lanewise`, `reduceLanes`, `compare`, `VectorMask.trueCount/toLong`, `compress`,
`rearrange`, `BIT_COUNT`) is present on 21.0.12 (`javap`) and survives into the finalized shape (*from memory*, to be
re-checked against the JDK 25 javadoc before the switch). Intrinsic availability per CPU was read from the match rules
in the installed `libjvm.so`; the target production CPU generation must be confirmed with `-XX:+PrintIntrinsics`.

## Gate

Builds and runs with and without the module; the provider's decision visible in the logs; the differential test
harness in place before the first kernel lands.

### LINKED ISSUE #1542: Roaring bitmap containers: fused AND + population-count kernel on `LongVector`
## Summary

First vector kernel (part of #1539; needs the provider issue): the 1024-word `bitCount(a[k] & b[k])` reduction that
sits behind every bitmap-container cardinality, and the count pass inside `BitmapContainer.and`.

## Where

`evita_roaring_bitmap/src/main/java/io/evitadb/roaringbitmap/BitmapContainer.java` — `andCardinality(BitmapContainer)`
(`:273-279`), the same loop shape at `:331`, `:388`, `:427`, `:911`; `and(BitmapContainer)` (`:221-235`, which runs the
count pass and then a second pass into the result); `Util.java:326-361` (`cardinalityInBitmap*Range`).
`FastAggregation.workShyAnd` (`FastAggregation.java:109-122`) uses the same word loops over its 1024-long buffer.

## Mechanism

```java
LongVector acc = LongVector.zero(S);
for (int k = 0; k < 1024; k += S.length()) {
    LongVector a = LongVector.fromArray(S, x, k);
    LongVector b = LongVector.fromArray(S, y, k);
    acc = acc.add(a.and(b).lanewise(VectorOperators.BIT_COUNT));
}
return (int) acc.reduceLanes(VectorOperators.ADD);
```

1024 is a multiple of every species length, so no tail. The fused variant also stores `a.and(b)` into the result
container, turning `BitmapContainer.and` from two passes into one. The same kernel gives `orCardinality`,
`xorCardinality`, `andNotCardinality` and `getCardinality` recomputation.

## Expected effect and CPU dependence

With `vpopcntq` (AVX-512 VPOPCNTDQ) the loop is bound by two loads per step: roughly 3–6× over the scalar loop. On
AVX2-only x86, HotSpot 21 intrinsifies `BIT_COUNT` through a lookup-table path (`vpopcount_avx_reg`,
`StubRoutines::x86::_vector_popcount_lut`, verified in the installed `libjvm.so`), so the kernel still gains, estimated
1.5–2.5×. On NEON, roughly break-even to 1.5×. Whether C2 already auto-vectorizes this exact reduction on 21 is
unverified; check with `-XX:+PrintCompilation` on `BitmapContainer.andCardinality` first — if it does, the gain
shrinks to the wider-popcount advantage on AVX-512.

## Gate

`RoaringBitmapGroup` / `ArrayOrRoaringBitmap` spikes extended with an `andCardinality` case: ≥ 3× on the kernel on
AVX-512, a measured (not assumed) gain or a disabled kernel on AVX2, differential test green, the full roaring suite
green in both provider modes; then the facet states from the prerequisite issue again.

### LINKED ISSUE #1543: Roaring array containers: SIMD all-pairs intersection cardinality kernel (materializing variant as a stretch goal)
## Summary

Second vector kernel (part of #1539; needs the provider issue): intersection of two sorted 16-bit array containers,
cardinality first, materializing variant as a stretch goal.

## Where

`evita_roaring_bitmap/src/main/java/io/evitadb/roaringbitmap/Util.java` — `unsignedLocalIntersect2by2Cardinality`
(`:896-947`, count only), `unsignedLocalIntersect2by2` (`:1229-1290`, materializing), dispatched by
`unsignedIntersect2by2` (`:818-834`) to a galloping search when one side is 25× smaller. Callers:
`ArrayContainer.and` / `andCardinality` (`ArrayContainer.java:248-292`).

## Mechanism (all-pairs block intersection)

Load 8 (128-bit species) `char`s from each side with `ShortVector.fromCharArray`; compare block A against block B and
against each of the other 7 rotations of B (`b.rearrange(ROTATION[r])` with precomputed `VectorShuffle` constants);
OR the masks; `mask.trueCount()` is the number of matches in the block pair; advance the side whose last element is
smaller, both when equal — the scalar merge, 64 pairs at a time. Values are unsigned 16-bit: use `UNSIGNED_*`
comparisons for the advance test; equality is sign-agnostic. The galloping branch stays scalar; SIMD does not help a
binary search.

- **Count-only kernel:** needs only `compare`, `or`, `rearrange` (in-lane on a 128-bit species) and `trueCount`,
  all intrinsified on any AVX2 CPU and on NEON (`rearrangeS`, `rearrangeS_avx`, `rearrangeS_evex` rules present in
  the installed `libjvm.so`). Expected 2–4× over the scalar merge for inputs of comparable size.
- **Materializing variant (stretch):** writing matched lanes in order needs `compress`, which is intrinsified only on
  AVX-512 (VBMI2 for 16-bit lanes); elsewhere a 256-entry shuffle-table emulation per 8 lanes or widening to 32-bit
  lanes for `vpcompressd` on AVX-512F. Low priority; do only if the count kernel proves out and a profile names
  `ArrayContainer.and`.

## Gate

≥ 2× on the kernel for equal-size inputs of 256–4 096 elements; the full roaring test suite in both provider modes;
differential test on randomized inputs including all-equal, disjoint and one-sided-galloping shapes.

### LINKED ISSUE #1544: Roaring array-vs-bitmap probes: gather-based bit-test kernel (low priority)
## Summary

Low-priority vector kernel (part of #1539; needs the provider issue): the per-element bit probes that intersect an
array container with a bitmap container.

## Where

`evita_roaring_bitmap/src/main/java/io/evitadb/roaringbitmap/BitmapContainer.java` — `and(ArrayContainer)`
(`:200-209`) and `andCardinality(ArrayContainer)` (`:256-264`), both probing `bitValue(v)` =
`(bitmap[v >>> 6] >>> v) & 1` per array element (`:580-582`).

## Mechanism

Widen 8 `char`s to an `IntVector`, compute the word indices (`v >>> 6`) into an `int[]` index map, gather the words
with `LongVector.fromArray(S, bitmap, 0, idx, 0)`, shift each lane by `v & 63` (per-lane variable shift, `vpsrlvq`),
AND with 1, sum with `reduceLanes(ADD)` (count) or use the resulting mask to append matches (materializing, `compress`
on AVX-512 only).

## Why low priority

Gathers run at roughly one lane per cycle on current x86, so the ceiling is 1.5–2× over a scalar loop that already has
no branches. Do it only if a profile names these two methods after the bitmap and array kernels landed.

## Gate

Measured on the `ArrayOrRoaringBitmap` spike with mixed container pairs; differential test; the roaring suite in both
provider modes.

novoj:
Builds, measures and where necessary rejects the four SIMD proposals the feasibility record
`2026-09-10-simd-vector-api-feasibility` made for the vendored roaring bitmap. Two of the four ship; two are
rejected with numbers. The campaign's biggest measured win turned out not to be a vector kernel at all, but a
scalar policy borrowed from CRoaring.

Full reasoning, every rejected option and its measurements:
`documentation/adr/2026-09-19-roaring-simd-kernels-and-lazy-union/README.md`.

## What ships

- **Optional Vector API provider** (`requires static jdk.incubator.vector`) with a scalar twin per kernel,
  self-detection of the module, the JIT and the CPU at class-init, per-kernel switches, a self-test against the
  scalar twin on all three output aliasings, and both-mode surefire runs. A missing module, a C1-only JIT or a
  narrow CPU falls back to scalar rather than failing. The javac flag is scoped to the roaring POM so the
  incubator module stays unresolvable inside the JDK 17 driver modules.
- **Fused bitmap word kernels** — single-pass store + count, replacing the count-then-store pairs. 1.6–2.0x on
  `iand`, 1.35x on `or`, and the *scalar* fused loop captures nearly all of it.
- **A CRoaring-style lazy array union on the multi-way path**, with its own bound (64 values) and an input-count
  guard. This is the change that reaches the profile.
- **Sparse-aware extraction** that skips all-zero 8-word blocks, behind the provider.
- Two engine-side changes folded in: facet ids grouped by group id instead of by hashing the index object
  (5.4 % of facet-query CPU), and `ArrayContainer`/`RunContainer` hash codes computed in O(1) from the last
  seven entries — the value is unchanged, and a 20,000-container differential confirms it.
- **Three copy-on-write defects fixed**, found by a dedicated audit of the sharing layer and a follow-up
  review pass: `priorityqueue_or` handing out containers its inputs still own (introduced by evita's borrowing
  layer, where upstream clones); an emptied chunk letting `xor` return its own argument (inherited from
  upstream, byte-identical in 1.6.18 — worth reporting there); and the same degenerate short-circuit surviving
  in `andNot`, where `run.andNot(emptyArray)` returned its own receiver. None is reachable from engine code
  today — a `RoaringArray` never stores an empty container, and `priorityqueue_or` has no engine caller — but
  the pattern is now established three times, so each is fixed rather than noted. Every fix carries a
  regression test proven to fail without it.

  The `andNot` case is worth calling out as a lesson about test shape: `ContainerBinaryOpFreshnessTest` exists
  precisely to prove out-of-place operators never alias their operands, and it passed throughout — because none
  of its fixtures was ever empty, and an empty operand is the only input that reaches the short-circuit. It now
  sweeps empty operands across every shape, operator and operand order.

## What is rejected, and why

- **#1543, all-pairs 8x8 block intersection — 0.5–0.6x the scalar merge** at every size and density. Without
  `pcmpestrm` (not expressible in the Vector API) a block costs 8 compares + 7 rotations + 7 mask ORs. An
  unrolled variant recovers 1.2–1.5x at >= 1024 equal-size elements, but both operands exceed 257 values in only
  4.4 % of production intersections.
- **#1544, gather-based probe — 0.89x** on 900 real probe pairs. The scalar bit test already runs at the
  load-throughput bound.
- **Vector `andCardinality` and siblings — 0.85x.** C2 on 21.0.12 already auto-vectorises the
  `bitCount(a & b)` reduction; with `-XX:-UseSuperWord` the scalar loop is 1.7x slower. The feasibility
  analysis had marked this hypothesis unverified and estimated 3–6x against a scalar loop that does not exist
  on this JIT.
- **A word-batched array-to-bitmap scatter**: 1.42x on a value-weighted replay, **+7.3 % end to end**. Caught on
  a quiet box before merge and reverted; the median real operation scatters 4 values into 4 distinct words,
  where batching saves no write and adds a branch per value.

## Why the lazy union is the real change

A census of 52.5 M container operations from 2,600 facet, hierarchy, price and attribute queries against a
production e-commerce catalog (475 k entities, 565 k entity indexes) found 92.7 % of them to be a scatter of a
median-4-value array into an 8 KiB bitmap, and bitmap x bitmap pairs to be 0.23 %. 95.3 % of the 1.93 M bitmaps
built by lazy unions demote straight back to an array at repair time, and those promotions were 23 % of every
byte the workload allocated. The workload is not word arithmetic, so word kernels could not have paid.

## Verification

- Roaring module: 18,035 tests, 0 failures, in **both** surefire executions (vector kernels and scalar kernels).
- Differential tests scalar-versus-vector for every kernel, plus a lazy-cardinality protocol test.
- End to end on a serialized 2,600-query replay against the production catalog, quiet box, interleaved arms,
  38,842,917 records matched and verified identical across builds: **facet summaries -25.4 %, facet+hierarchy
  summaries -35.1 %**, catalog-wide facet summary +1.7 % (ranges overlap), total -1.9 %.

Closes #1541
Closes #1542
Closes #1543
Closes #1544

#1539 stays open: the umbrella still needs #1540 and #1545.


novoj:
**Review findings addressed** — `d6f9e74e1`, `84428d561`.

### MAJOR — absolute paths into a personal worktree

Valid, and worse than a portability wrinkle: the path pointed inside a git-ignored `specifications/` folder that leaves the disk when its work finalizes, so the corpus it named is about to stop existing even on the machine that wrote it.

All three now take the dump as a `@Param` defaulting to `<path-to>/operands.bin` / `<path-to>/unions.bin` — a value that cannot accidentally resolve to a real file — and each "not readable" message now says the dump is captured from a running engine and is not carried in the tree, instead of only reporting that a file is missing.

`RoaringScatterShapeBenchmark` needed more than a new string: its path was a `private static final` read by three `static` helpers, so it is now an instance `@Param` threaded through them. That also puts it in the result JSON and in the benchmark's own `# scatter:` describe line — the property `RoaringReplayBenchmark`'s javadoc already argues is the reason to prefer a parameter over a system property.

One correction to the review: `RoaringHashBenchmark` was cited as the class that takes the path as a bare `@Param`, but its `@Param` values are container-size labels (`c4` … `c4096`), not paths. The class that does what you described is `RoaringSparseExtractBenchmark`, which accepts a shape label and treats the value as a path only when it is neither `u<n>` nor `c<n>`. That is the pattern the three fixed classes now follow.

### NIT — fully-qualified `java.util.Arrays`

Fixed. Imported, and the three call sites shortened.

### NIT — the javadoc caveat on `allPairsTreeIntersect`

Right to flag, and it was worse than stale: **both** claims were false. I verified that by running the kernel rather than by reading it, since a comment that had already misled a reviewer was not something to correct by reasoning.

The two kernels perform byte-identical stores — `va.compress(found).intoCharArray(out, outPos, VectorMask.fromLong(SHORT_128, (1L << matches) - 1L))`. That store is lane-masked to `matches`, so it writes exactly that many entries and never a whole vector.

Measured, both arms against a scalar merge:

- 20,000 random operand pairs with a destination of exactly `min(la, lb)` — no overflow, results identical, zero failures.
- The densest possible shape (`a == b`, so every lane matches and the output cursor tracks the read cursor) at every width from 1 to 4,096, the full `ArrayContainer` range — clean with exact slack **and** fully aliased (`out == a`).

So the `+ 8` slack was unnecessary and the "not alias-safe" warning was simply wrong. That mattered more than a nit: the all-pairs mechanism was rejected on **measurement**, and a false structural blocker sitting beside it reads as a second, independent reason it can never ship. Both the javadoc and the same sentence in the record's evidence file (`benchmarks/REPORT.md`) are corrected.

The three `Not alias-safe` notes on the *oriented broadcast probe* are a different and correct argument — orientation can make the walked array and the destination different arrays — and are untouched.

PR #1610, opened 2026-09-19
fix: stop facet statistics depth NONE running the IMPACT path

### LINKED ISSUE #1609: facetSummary depth NONE runs the IMPACT path although it is documented as the cheapest option
## Summary

`FacetStatisticsDepth` documents its three values as a cost ladder — *"from cheapest to most expensive"* — with
`NONE` described as *"no facet statistics are computed. The engine returns the list of facet options … without any
counts or impact. This is the cheapest option."*

The engine gates on the middle rung instead, so `NONE` lands on the expensive branch together with `IMPACT`.

## Where

`ReferenceSummaryProducer$GroupAccumulator` constructor, `ReferenceSummaryProducer.java:1144`:

```java
this.impactCalculator = referenceSummaryRequest.facetStatisticsDepth() == FacetStatisticsDepth.COUNTS ?
    ImpactCalculator.NO_IMPACT : impactCalculator;
```

| depth | gets impact | should |
|---|---|---|
| `COUNTS` | no | correct |
| `IMPACT` | yes | correct |
| `NONE` | **yes** | **no** |

`FacetStatisticsDepth.NONE` occurs only twice in `evita_engine`, both in `ReferenceSummaryProducer` — line 490
(dropping facet-emitted groups for histogram-only references, *after* they have been computed) and a javadoc
mention at line 547. Neither gates any computation.

## Counts are computed unconditionally too

`FacetAccumulator.toFacetStatistics` (`:1258`) always calls `getCount()`, which builds an intersection of the
query baseline with the union of the group's facet bitmaps and reads its size. No depth check precedes it.

This half is not a one-line fix: `FacetStatistics.count` is a primitive `int`
(`ReferenceSummary.java:439`), so "not computed" cannot be expressed — only `0`, which is indistinguishable from
a facet that genuinely matches nothing. Honouring the documented contract for counts therefore means either
making the field nullable (an API and serialization change across gRPC, GraphQL and REST) or amending the enum
documentation to say that `NONE` still returns counts and only skips impact. **That decision is open.**

## Impact

A caller asking for the cheapest option receives the most expensive computation available on the facet path: a
separate hypothetical result-set evaluation per unselected facet option. Correctness of returned data is not
affected — the values are right, they are merely computed when they were not asked for.

## Suggested fix

Gate on the value that wants the work rather than the one that does not:

```java
this.impactCalculator = referenceSummaryRequest.facetStatisticsDepth() == FacetStatisticsDepth.IMPACT ?
    impactCalculator : ImpactCalculator.NO_IMPACT;
```

Then decide the counts question separately.

## Gate

A functional test asserting that every `FacetStatistics.getImpact()` is `null` at depth `NONE`; it fails against
the current code because impact is genuinely computed. Note that no existing test covers depth `NONE` at all.

novoj:
`FacetStatisticsDepth` documents its three values as a cost ladder — *"from cheapest to most expensive"* — with
`NONE` returning the facet options *"without any counts or impact"*. The engine has a single place where that
depth switches work off, and it tested for the middle rung:

```java
this.impactCalculator = referenceSummaryRequest.facetStatisticsDepth() == FacetStatisticsDepth.COUNTS ?
    ImpactCalculator.NO_IMPACT : impactCalculator;
```

`COUNTS` therefore skipped impact correctly, while `NONE` fell through to the same branch as `IMPACT`. A caller
asking for the cheapest depth received the most expensive computation on the facet path — a hypothetical
result-set evaluation per facet option.

| depth | got impact | should |
|---|---|---|
| `COUNTS` | no | correct |
| `IMPACT` | yes | correct |
| `NONE` | **yes** | **no** |

Gating on `IMPACT` instead gives every other depth the null calculator, which is what the ladder describes.

No returned data was wrong — the values were correct, they were merely computed when they had not been asked
for.

## What this does not change

**Counts are still computed at every depth, `NONE` included.** `FacetStatistics.count` is a primitive `int`, so
"not computed" cannot be expressed other than as `0`, which is indistinguishable from a facet that genuinely
matches nothing and would silently mislead any client rendering it. Honouring the documented contract for counts
needs either a nullable field — an API and serialization change across gRPC, GraphQL and REST — or a correction
to the enum's own documentation. That is a design decision and it is recorded on the issue rather than taken
here.

## Verification

`FacetStatisticsDepthCostLadderTest` is a new self-contained functional test: it builds its own catalog of nine
products across three brands on a faceted reference and asserts impact nullity at each of the three depths. It
observes the routing decision through the nullity of the reported impact rather than through the numbers, since
the null calculator returns `null` for every facet while the real one returns a populated impact.

Depth `IMPACT` is covered as a **positive control**, not as extra coverage: two null-assertions alone would also
pass against a fixture that produced no facets at all, so the third case is what proves the other two are not
vacuously satisfied. A guard inside the shared helper fails loudly if a query yields no facet statistics, for
the same reason.

With the gate reverted, exactly one of the three fails:

```
FacetStatisticsDepthCostLadderTest.shouldNotComputeImpactAtDepthNone
  facet 1 carries an impact at depth NONE, which documents that neither counts nor impact are computed
  and is the cheapest depth ==> expected: <null> but was: <-6>
```

The other two staying green is the point — `COUNTS` remains suppressed and `IMPACT` remains computed — so the
test moves only with the routing decision. Built with `-am`, so the run tested the modified engine rather than a
stale artifact.

Closes #1609


novoj:
**Review nit addressed in #1613**, since this merged before it could land here.

`FacetStatisticsDepthCostLadderTest`'s three tests are grouped under `@Nested class ImpactRouting`, matching `.claude/rules/testing.md` and the neighbouring `ReferenceSummaryScopeLeakTest` in the same package. The positive control stays beside the two assertions it validates rather than being split off. Structure only — no assertion, fixture or tag changed; the three tests pass under the nested class.

PR #1612, opened 2026-09-19
fix: let the random query generator select every candidate it is given

### LINKED ISSUE #1611: RandomQueryGenerator.pickRandom can never select the first candidate and throws on a single-entry map
## Summary

`RandomQueryGenerator` picks a random candidate by walking a collection's iterator to a random index. The index
arithmetic is off by one in both overloads, so **the first element of every candidate collection is unreachable**,
and the `Map` overload throws for a single-entry map.

Every benchmark built on this generator inherits both — the artificial facet, hierarchy, price and attribute
states, and the `Client*` twins.

## Where

`evita_test/evita_performance_tests/src/main/java/io/evitadb/performance/generators/RandomQueryGenerator.java`

```java
// :89-97 — Set overload
final int index = theSet.size() == 1 ? 0 : random.nextInt(theSet.size() - 1) + 1;
final Iterator<T> it = theSet.iterator();
for (int i = 0; i < index; i++) { it.next(); }
return it.next();
```

For `size >= 2` the index lands in `[1, size - 1]`, and the walk returns the element at that position — so
position 0 is never returned. Only `size - 1` of the `size` candidates can ever be selected.

```java
// :100-107 — Map overload
final int index = random.nextInt(filterableAttributes.size() - 1) + 1;
```

Same off-by-one, plus no `size == 1` guard: `nextInt(0)` throws `IllegalArgumentException`, so a schema with
exactly one filterable attribute crashes the generator rather than selecting it.

The same body is copy-pasted into two long-running tests, with the same defect:

- `evita_test/evita_long_running_tests/src/test/java/io/evitadb/api/LongRunningEvitaGenerationalTest.java`
- `evita_test/evita_long_running_tests/src/test/java/io/evitadb/api/LongRunningEvitaReferencesGenerationalTest.java`

## Related off-by-one, same family

`:212` — `histogramCount = size == 1 ? 1 : 1 + random.nextInt(size - 1)` never reaches `size`, so a histogram
query can never span all numeric attributes.

Note the coupling: the rejection loop at `:216-219` draws until it finds an unselected attribute. Because
`pickRandom` can only reach `size - 1` distinct values today, that loop terminates only because the count is
capped one short as well. Fixing `pickRandom` first makes the loop strictly safer; fixing `:212` alone would
make it spin forever.

## Impact

Measurement coverage, not correctness of the engine. One attribute, one faceted reference, one hierarchy and one
price list — whichever sorts first in each collection's iteration order — are never exercised by any benchmark
built on this generator. A single-attribute schema cannot be benchmarked at all.

## Not addressed here

The generator's output is **not reproducible across JVMs despite a fixed seed** — recorded in
`documentation/adr/2026-09-19-roaring-simd-kernels-and-lazy-union/README.md`, which observed three distinct
query checksums and two distinct record totals for one seed. `pickRandom`'s dependence on iteration order is the
obvious suspect, but the cause has not been diagnosed and should not be assumed. That is its own task; comparing
two builds today requires replaying a serialized query file.

## Gate

A unit test over candidate sets of sizes 1, 2 and 5 asserting that every element is reachable across enough
draws, and that a single-entry map returns its only entry rather than throwing.

novoj:
`RandomQueryGenerator.pickRandom` chooses a candidate by walking a collection's iterator to a random index, and
both overloads drew that index from `[1, size)` instead of `[0, size)`. The walk returns the element at that
position, so **whichever candidate a collection iterates first could never be selected** — one attribute, one
faceted reference, one hierarchy and one price list were invisible to every benchmark built on this generator.

The `Map` overload also lacked the empty guard its sibling carries, so a schema with exactly one filterable
attribute hit `nextInt(0)` and threw rather than selecting it.

The same body is copy-pasted into `LongRunningEvitaGenerationalTest` and
`LongRunningEvitaReferencesGenerationalTest`, and is corrected there too.

## The coupled off-by-one

`generateRandomAttributeHistogramQuery`'s count drew from the same short range, so a histogram query could never
span every numeric attribute. Fixing it is only safe together with the above: the rejection loop below it draws
until it finds an unselected candidate, and while `pickRandom` could reach only `size - 1` distinct values that
loop terminated purely because the count also stopped a step short. With every candidate reachable, the loop can
satisfy the full span.

## Impact

Measurement coverage, not engine correctness. Nothing about stored or returned data changes — the benchmarks
simply stop ignoring part of the schema they were pointed at. Historical benchmark numbers shift, which costs
nothing here: the production datasets these run against are restored from daily backups and never come back
identical, so figures were only ever comparable within a single snapshot.

## Not addressed

The generator's output is **not reproducible across JVMs for a fixed seed**, recorded in
`documentation/adr/2026-09-19-roaring-simd-kernels-and-lazy-union/README.md`, which observed three distinct query
checksums and two distinct record totals from one seed. `pickRandom`'s dependence on iteration order is the
obvious suspect, and deliberately not assumed to be the cause. Comparing two builds still requires replaying a
serialized query file.

## Verification

`RandomQueryGeneratorTest` exercises the set-based selection through the histogram generator, the surface that
takes a candidate set directly. With the fix reverted:

```
shouldReachEveryCandidateIncludingTheFirst
  `alpha` is iterated first and was never selected across 500 draws, so the index the selection walks to
  cannot reach position zero ==> expected: <true> but was: <false>
shouldBeAbleToSpanEveryCandidate
  the widest histogram requested 2 of 3 candidates, so a query spanning all of them is unreachable
  ==> expected: <3> but was: <2>
```

**Two of the three tests move with this change, not three.** The map-based selection is the same arithmetic but
is verified by reading rather than by test — reaching it needs an entity schema with populated attribute
statistics — and `shouldSelectTheOnlyCandidate` pins behaviour the set overload already had. Both limits are
documented at the tests themselves rather than left for a reader to discover.

Closes #1611


novoj:
**Review: no findings to address.** Recording why this merged with a red build, since the check will stay red in the history.

The failure was a single test, `EvitaTest.shouldCreateAndRenameCollectionInTransaction` (`expected: <false> but was: <true>`), and it cannot be reached by this change:

- The assertion at `EvitaTest.java:1666-1673` spins for a bounded 2,000 ms waiting for an asynchronous collection-file deletion, then asserts the file is gone. On a loaded runner the cleanup misses that window.
- This PR touches only `evita_test/evita_performance_tests` and `evita_test/evita_long_running_tests`. `evita_functional_tests`, where the failing test lives, declares no dependency on either — there is no path by which the diff reaches it.
- The run totalled 24,372 tests, the same as the contemporaneous run on #1610, which passed.

A re-run of the failed jobs was started to confirm. Nothing in this PR is implicated either way.

novoj:
Re-run of the failed jobs completed: **success**. The failure was the flake described above, not this change.

